PR #50, opened 2023-05-24
Helm chart

runyontr:


kubernetes/observibility/dcgm-exporter.yaml
```diff
@@ -0,0 +1,74 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
```
runyontr, line 1: We'll bring this stuff into the chart sometime soon

PR #125, opened 2023-06-29
feat: add faster-whisper impl with new gRPC spec

DayOfThePenguin:
Rebased my faster-whisper implementation to use the go api backend and the new gRPC spec

### DIFF
--- a/api/backends/hf/hf.go
+++ b/api/backends/hf/hf.go
@@ -1,0 +1,155 @@
+package hf
+
+import (
+	"context"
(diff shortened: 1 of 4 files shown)


DayOfThePenguin:
It might be worth me adding an example of how to run the faster-whisper model container locally with GPUs before we merge this so it's easier to reproduce

DayOfThePenguin:
`docker run -it -p 50051:50051 --gpus all leapfrogai/faster-whisper:latest`

PR #138, opened 2023-07-10
Add config endpoint and example to mpt-7b-chat

gerred:


gerred:
keeping this open while we merge this in with another branch, but this PR itself will end up getting closed.

PR #154, opened 2023-07-31
Add API auth with optional postgres DB

DayOfThePenguin:
Existing leapfrog functionality is unchanged if you add the -nodb flag when you run the binary.

Database stand-up and initial migrations are in /migrations. docker-compose demonstrates the stand-alone job for db initialization and migration.

Also added in `finish_reason` handling for compatibility with smart-chatbotui and the other chatbotui clones.


gerred:
After talking with everyone, going a different direction to keep auth itself out of LeapfrogAI.

PR #259, opened 2024-02-28
Migrate Hugo Docs into main repo

gphorvath:


jalling97:
Does it make sense to merge the `docs` and `docs-website` folders? It looks like the only usage of the current `docs` folder is to hold the vectordb ADR, a leapfrogai png, and the customer pngs. I feel like the pngs could be moved to the assets subfolder and the ADRs remain its own folder within the combined `docs` or moved to a wiki, however we want to track those going forward.

jalling97 (approved):

bdfinst (approved):

PR #268, opened 2024-03-19
chore: Consolidate our repos into a monorepo

YrrepNoj:
Consolidate several of our repo's into a monorepo.

This PR consolidates the functional bits from the following repositories:
  - https://github.com/defenseunicorns/leapfrogai-api
  - https://github.com/defenseunicorns/leapfrogai-sdk

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,13 +27,5 @@ config.json
 generation_config.json
 vocabulary.json
 
-
-# Generated Code
-src/leapfrogai_api/types/audio
-src/leapfrogai_api/types/chat
-src/leapfrogai_api/types/completion
-src/leapfrogai_api/types/embeddings
-src/leapfrogai_api/types/name
-
 # go binaries
 api/main
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ temp/
 zarf-sbom/
 /values.yaml
 .pytest_cache
+*egg-info
+build/
+*.whl
 
 # local model and tokenizer files
 *.bin
@@ -23,6 +26,14 @@ tokenizer.json
 config.json
 generation_config.json
 vocabulary.json
+
+
+# Generated Code
+src/leapfrogai_api/types/audio
+src/leapfrogai_api/types/chat
+src/leapfrogai_api/types/completion
+src/leapfrogai_api/types/embeddings
+src/leapfrogai_api/types/name
 
 # go binaries
 api/main
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,13 @@ help: ## Display this help information
 		{printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 
+clean: ## Clean up all the things (packages, build dirs, compiled .whl files, python eggs)
+	-rm zarf-package-*.tar.zst
+	-rm -rf build
+	find . -name '*.whl' -delete
+	find . -name '*.egg-info' -type d -exec rm -rf {} +
+
+
 build-wheel: ## Build the wheel for the leapfrogai_api module
 	python -m pip wheel . -w build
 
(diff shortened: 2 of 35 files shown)


.gitignore
```diff
@@ -24,5 +27,13 @@ config.json
 generation_config.json
 vocabulary.json
 
+
+# Generated Code
+src/leapfrogai_api/types/audio
```
YrrepNoj, line 32: At first I thought it would make sense to not check in the generated code. I started second guessing that over the weekend though...

Any thoughts?

Makefile
```diff
@@ -0,0 +1,44 @@
...
+			--python_out=src/. \
+			--grpc_python_out=src/. \
+			src/leapfrogai_api/types/proto/leapfrogai_api/types/**/*.proto
+
+
+local-registry: ## Start up a local container registry. Errors in this target are ignored.
+	-docker run -d -p 5000:5000 --restart=always --name registry registry:2
+
+
+build-api: local-registry build-wheel ## Build the leapfrogai_api container and Zarf package
```
YrrepNoj, line 39: This make target is a bit... ✨ extra ✨. It's a little complex, but I think it achieves a good thing...

Any thoughts?

packages/api/zarf.yaml
```diff
@@ -0,0 +1,29 @@
...
+    charts:
+    - name: leapfrogai 
+      namespace: leapfrogai
+      localPath: chart
+      version: 0.0.1
+      valuesFiles:
+        - "lfai-values.yaml"
+    images:
+      - "ghcr.io/defenseunicorns/leapfrogai/api:###ZARF_PKG_TMPL_LEAPFROGAI_IMAGE_VERSION###"
+      - "kiwigrid/k8s-sidecar:1.23.3"
```
YrrepNoj, line 29: NOTE: We switched to dockerhub for our kiwigrid image.

pyproject.toml
```diff
@@ -0,0 +1,76 @@
+[project]
```
YrrepNoj, line 1: @gerred You and I should talk about how we want to tackle this pyproject.toml file.

src/leapfrogai_api/main.py
```diff
@@ -0,0 +1,39 @@
+import asyncio
```
YrrepNoj, line 1: Add the [unit tests we already had](https://github.com/defenseunicorns/leapfrogai-api/blob/main/tests/test_main.py) for the api python code

packages/api/zarf.yaml
```diff
@@ -0,0 +1,29 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/defenseunicorns/zarf/main/zarf.schema.json
```
YrrepNoj, line 1: Add an [e2e test workflow](https://github.com/defenseunicorns/leapfrogai-api/blob/main/.github/workflows/test-deploy.yaml) that validates this package can be deployed.

packages/api/chart/values.yaml
```diff
@@ -0,0 +1,10 @@
+image: 
+  lfaiAPITag: v0.5.1
```
YrrepNoj, line 2: NOTE: The default image version will need to change in the near future (as we synchronize all of our version numbers)

.gitignore
gphorvath: Maybe we can just start by checking it in follow the same pattern as the source repo? We can always move away from this practice in subsequent PRs if we decide we don't like it.

packages/api/zarf.yaml
gphorvath: 🥳

Makefile
```diff
@@ -0,0 +1,44 @@
+ARCH ?= amd64
+KEY ?= ""
+
+VERSION ?= $(shell git describe --abbrev=0 --tags)
+LOCAL_VERSION ?= $(shell git rev-parse --short HEAD)
+
+######################################################################################
+
+.PHONY: help
+help: ## Display this help information
```
jalling97, line 10: Would it make sense to put back some of the venv make targets for simplicity?
```suggestion
help: ## Display this help information
create-venv:
	python -m venv .venv
requirements: ##or dependencies or something
        python -m pip install .
```
Alternatively, some instructions to set up a virtual environment in the README, because at least one of the make targets won't work without installing at least some of the dependencies in `pyproject.toml`

Makefile
jalling97: I like that it's a one stop shop for building the zarf package. It's nice to not have to fish for the local docker registry commands. Is it worth having a make target for deploying the api via the package just created? (or instructions via the README?)

andrewrisse:
no comment other than I'm ready to move the new frontend in to the new monorepo anytime

.gitignore
YrrepNoj: I agree, for now at least. 
- [ ]  remove this from the .gitignore

gerred: gRPC / proto code is meant to be checked in and versioned. You'd do the same with generated code from go generate.

Makefile
YrrepNoj: We 10,000% need to add more details to the README.


We sure could have a make target for deploying. Deploying the packages isn't as complex as building the packages so the make targets will be fairly trivial. 👍 

I am going to resolve this for now, and add issues to add these improvements in the near future.

Makefile
YrrepNoj: I think not, at least for now?
I am all for adding more instructions to the README instead of adding these to make targets though.


The one edge case for managing our python dependencies in the make targets could be for the `gen-python` target. If that is something that comes up in the future I'm down add something like this to that specific target.

gerred (approved):

PR #274, opened 2024-03-21
refactor: Consolidate vllm into the monorepo

YrrepNoj:
This PR consolidates the functional bits from the following repositories:

https://github.com/defenseunicorns/leapfrogai-backend-vllm


gphorvath:
Does this mean we won't be separately maintaining the model skeleton?

Makefile
```diff
@@ -64,3 +64,23 @@ build-llama: local-registry build-wheel ## Build the llama (cpu) container and Z
 
 	## Build the Zarf package
 	zarf package create packages/llama --registry-override=ghcr.io=localhost:5000 --insecure --set IMAGE_VERSION=${LOCAL_VERSION} --confirm
+
+
+build-vllm: local-registry build-wheel
+	## Download the wheels for the optional 'vllm' dependencies
+	pip wheel ".[vllm]" -w build
```
jalling97, line 71: @YrrepNoj using `pip wheel` builds all the dependencies as `.whl` files, circumventing the `tar.gz` problem. Let me know if that doesn't quite cover our bases.

This brings up the point, technically the `build-wheel` target being run during `build-vllm` is redundant since all of the primary dependencies get downloaded in the first command. It slows down the build process by like 10 seconds so not a big deal if keeping it there cements the fact that the wheels have to be downloaded.

YrrepNoj: Thanks for looking into that! I'll take a look to see if it does everything I expect too.

PR #315, opened 2024-03-25
test: add workflows for e2e tests, pytest, and release publishing

YrrepNoj:
This PR introduces several PRs.
 - e2e tests that build, deploy, and test our API & various model backends
 - pytest that runs against the leapfrogai_api code
 - a release workflow that creates and publishes the relevant container images and zarf packages.


gerred (approved):

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,112 @@
+name: e2e
+on:
+  pull_request:
+    paths:
```
justinthelaw, line 4: Is there a reason this isn't just `paths-ignore`, like the example below?

```yaml
  pull_request:
    paths-ignore:
      - "**.md"
      - "docs/**"
      - "CODEOWNERS"
```

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,112 @@
...
+        - name: Test text-embeddings
+          run: |
+            python -m pytest ./tests/e2e/test_text_embeddings.py -v
+
+        - name: Cleanup text-embeddings
+          run: |
+            uds zarf package remove text-embeddings --confirm
+
+        ##########
+        # whipser
```
justinthelaw, line 100: Spelling haha!

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,112 @@
...
+    runs-on: ai-ubuntu-big-boy-8-core
+
+    steps:
+        - name: Checkout Repo
+          uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+        - name: Setup Python
+          uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c #v5.0.0
+          with:
+            python-version: '3.11.6'
```
justinthelaw, line 38: I still (sort of) think we should have this version in the `project.toml` and then pin the python version using that in this action. Example:

```yaml
- name: Set up Python
        uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c # v5.0.0
        with:
          python-version-file: 'pyproject.toml' # Read python version from a file pyproject.toml
```

I know @gerred had said there was a reason to be a bit more loose with the python version, but if everything theoretically passes in this workflow then we should keep it at 3.11.6. OR, we should have a matrix to test other python versions.

.github/workflows/pytest.yaml
```diff
@@ -0,0 +1,44 @@
...
+  pytest:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c #v5.0.0
+        with:
+          python-version-file: 'pyproject.toml'
```
justinthelaw, line 32: Inconsistency with my previous comment on version pin.

.github/workflows/pytest.yaml
```diff
@@ -0,0 +1,44 @@
+name: pytest
+on:
+  pull_request:
+    paths-ignore:
```
justinthelaw, line 4: Inconsistency with my previous comment on `paths-ignore`

.github/workflows/release.yaml
```diff
@@ -0,0 +1,102 @@
...
+          zarf package create . --set=LEAPFROGAI_IMAGE_VERSION=${{ steps.get_version.outputs.version-without-v }} --architecture arm64 --confirm
+
+          zarf package publish zarf-package-leapfrogai-api-amd64-${{ steps.get_version.outputs.version-without-v }}.tar.zst oci://ghcr.io/defenseunicorns/packages/leapfrogai
+          zarf package publish zarf-package-leapfrogai-api-arm64-${{ steps.get_version.outputs.version-without-v }}.tar.zst oci://ghcr.io/defenseunicorns/packages/leapfrogai
+
+          rm zarf-package-leapfrogai-api-*.tar.zst
+
+      - name: Build and Publish llama
+      # TODO: Is this the name we want for the image?
+      # TODO: HERE AND ELSWHERE: PACKAGE_VERSION is not used in the new zarf package config. Maybe just update the zarf.yaml IMAGE_VERSION to have the version metadata set by
```
justinthelaw, line 58: The package version variable was for ease of dev deployments and should be removed in our manifests. @CollectiveUnicorn has more context.

justinthelaw (commented):
Seems like this is still a WIP, but had some comments after a cursory look

.github/workflows/e2e.yaml
YrrepNoj: I should have left a comment on this! That is a very good point and something I agree with.

This is just temporary while I experiment with some things 😬

.github/workflows/pytest.yaml
YrrepNoj: I agree that this approach will be how we define the python versions for our workflows. Resolving as this is a no-op.

.github/workflows/e2e.yaml
YrrepNoj: `paths-ignore` does not allow the `!` operator but `paths` does. I was thinking about using `paths` for this workflow because this workflow is large & slow and I wanted to try to restrict when it would run with a little more granularity- avoiding unnecessary runs such as when a file in the `.github/` directory was updated that wasn't this exact file.

The other workflow (pytest) is able to run on a regular runner and only takes a few seconds to complete so I was not as worried about being super restrictive on when that workflow runs.

Do you have ideas on what might be better?

justinthelaw: I was that you could use `paths-ignore`  like the way it's used in the other workflow I commented on. Basically, anything that changes besides specific files (the ones you marked with `!`) would trigger the workflow. It seems you used a `*` to everything, so I figured you could just do the opposite of what is done here.

packages/llama-cpp-python/Dockerfile
```diff
@@ -12,8 +12,9 @@ ARG REVISION=3f65d882253d1f15a113dabf473a7c02a004d2b5
 
 # TODO: Standardize on a better approach for how we download models into our images across our packages
 COPY scripts/model_download.py scripts/model_download.py
-RUN REPO_ID=${REPO_ID} FILENAME=${FILENAME} REVISION=${REVISION} python3.11 scripts/model_download.py
-RUN mv .model/*.gguf .model/model.gguf
+COPY build/*.gguf .model/
```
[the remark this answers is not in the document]
justinthelaw: I like this, as long as we promise not to push packages/images locally anymore, this shouldn't have us accidentally pushing something up with the wrong model, e.g., we tested a different model and that's what is in our `.model/` instead of our default.

.github/workflows/e2e.yaml
YrrepNoj: Seems like GitHub docs recommend this approach.

> If you want to both include and exclude path patterns for a single event, use the `paths` filter prefixed with the `!` character to indicate which paths should be excluded.

source: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths

packages/llama-cpp-python/Dockerfile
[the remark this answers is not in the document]
jalling97: I might be missing something, but doesn't `model_download.py` default to downloading to `.model`? I don't think I'm seeing why we're copying it from the `build/` directory if it's not likely to be there. Wouldn't the `COPY` command be more like:
```
COPY .model/*.gguf .model/
```

tests/e2e/test_llama.py
```diff
@@ -0,0 +1,57 @@
...
+# def test_completions():
+#     completion = client.completions.create(
+#         model=model_name,
+#         prompt="What is your name?",
+#     )
+
+#     assert completion.model == model_name
+#     assert len(completion.choices) == 1
+#     assert len(completion.choices[0].text) > 0
+#     assert len(completion.choices[0].text) < 500
```
jalling97: Do we delete if we're not testing completions? Or should we test completions since it's still in the API for now?

packages/llama-cpp-python/Dockerfile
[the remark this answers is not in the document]
YrrepNoj: @jalling97 good question. It's trying to copy from the `build/` directory in there is a `model.gguf` file there. 

If there is one there (because you put one there to speed up your build, great! If not, the `model_download.py` will be run and the model will be in the correct spot because, like you said, it defaults to the `.model/` directory.

Part of the reason why it i s copying from `build/*.gguf` is because if the `.model` directory doesn't exist when you try to do the COPY, the docker build will fail. We expect the build directory to exist because that's where we are storing our .whl's so we don't have to worry about that failing.

tests/e2e/test_llama.py
YrrepNoj: I think you maybe be reviewing an old commit. This commented-out code should be deleted.

packages/llama-cpp-python/Dockerfile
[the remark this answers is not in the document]
jalling97: Ah I see the approach now. Mind if we add a comment to make that a little clearer (unless I'm missing it)?

Something like:
```suggestion
# pre-downloading the model to the local build dir will make subsequent builds faster
COPY build/*.gguf .model/
```

.github/workflows/e2e.yaml
justinthelaw: Hm, I might be missing it, but I don't see where it says you should not use `paths-ignore`?

YrrepNoj: Maybe I am reading too much into it. 🤷‍♂️  Using `paths-ignore` with the `!` operator I think is what the quote above was referring to when saying

> If you want to both include and exclude path patterns for a single event ...

justinthelaw: Based on our conversation about the number of entries that both `paths-ignore` and `paths` would introduce, both are effectively the same, with `paths` most likely winning out over time. That being said, here are some additional paths we should ignore when using `paths`:

```yaml
  paths:
    # Catch-all
    - "*"
  
    # Ignore updates to the .github directory, unless it's this current file
    - "!.github/**"
    - ".github/workflows/e2e.yml"
  
    # Ignore docs and website things
    - "!**.md"
    - "!docs/**"
    - "!website/**"
    - "!adr/**" # <- just added this one
    - "!netlify.toml" # <- just added this one
  
    # Ignore updates to generic github metadata files
    - "!CODEOWNERS"
    - "!.gitignore"
    - "!LICENSE" # <- just added this one
    # removed CONTRIBUTING.md since it's covered by "!**.md"
  
    # Ignore local development files
    - "!pre-commit-config.yaml" # <- just added this one
    
    # Ignore unit tests
    - "!tests/pytest/**" # <- just added this one
```

Anywhere is says "# <-  just added this one" or a comment about removal, just make sure to take it out before insertion!

YrrepNoj: @justinthelaw I took all of your recommendations other than `!adr/**` since those are captured by the `!**.md` rule anyways.

gerred (approved):
lgtm with comments resolved

jalling97 (approved):

justinthelaw (approved):

gphorvath (approved):

gerred (approved):

PR #322, opened 2024-03-26
Monorepo local UDS deployment and README Updates

jalling97:
Since this feature branch was originally branched off from `consolidate-workflows`, there are a few updates that are also contained in that branch, mostly in the `Makefile`.

Changes specific to this branch:
- Adds uds bundles specifically for deploying locally created zarf packages
- Updates the README to direct users to deploy via the standard uds deployments or how to build the zarf packages locally
- Adds instructions to the README for running each of the components in a `dev` context locally without K8s
- Makes minor updates to the docs site now that uds bundles are stored within the monorepo

Actions after merge:
- [UDS-LeapfrogAI](https://github.com/defenseunicorns/uds-leapfrogai) can be archived

PR #328, opened 2024-03-28
UDS Documentation Updates

jalling97:
Quick PR to update some the documentation:
- Point people to use UDS to deploy LFAI
- Deprecate instructions for k8s deployment that doesn't use UDS
- Minor other doc fixes

The plan is to adjust this documentation (and add more) once #322 gets finalized, but this documentation is up to date at the moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,15 +37,11 @@ Large Language Models (LLMs) are a powerful resource for AI-driven decision maki
 
 ### UDS
 
-UDS is the preferrred method for a full local deployment of LeapfrogAI. Instructions can be found on the [LeapfrogAI Documentation Site](https://docs.leapfrog.ai/docs/).
+The preferred method for running LeapfrogAI is a local [Kubernetes](https://kubernetes.io/) deployment using [UDS](https://github.com/defenseunicorns/uds-core). Simple instructions for this type of deployment can be found on the [LeapfrogAI Documentation Site](https://docs.leapfrog.ai/docs/).
 
 ### Tadpole
 
-> GitHub Repo:
->
-> - [Tadpole](https://github.com/defenseunicorns/tadpole)
-
-Tadpole is a simple way to get LeapfrogAI up and running in a sandbox environment. While not intended for production, it helps the user to understand the various components of LeapfrogAI and how they interact.
+[Tadpole](https://github.com/defenseunicorns/tadpole) is a simple, lightweight way to get LeapfrogAI up and running in a sandbox environment using Docker. While not intended for production, it helps the user to understand the various components of LeapfrogAI and how they interact.
 
 ## Components
 
--- a/README.md
+++ b/README.md
@@ -6,14 +6,15 @@
 - [Table of Contents](#table-of-contents)
 - [Overview](#overview)
 - [Why Host Your Own LLM?](#why-host-your-own-llm)
+- [Getting Started](#getting-started)
+  - [UDS](#uds)
+  - [Tadpole](#tadpole)
 - [Components](#components)
-  - [Getting Started](#getting-started)
   - [API](#api)
   - [Backends](#backends)
   - [Image Hardening](#image-hardening)
   - [SDK](#sdk)
   - [User Interface](#user-interface)
-  - [Advanced Deployments](#advanced-deployments)
 - [Community](#community)
 
 ## Overview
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -32,34 +33,36 @@ Large Language Models (LLMs) are a powerful resource for AI-driven decision maki
 
 - **Mission Integration**: By hosting your own LLM, you have the ability to customize the model's parameters, training data, and more, tailoring the AI to your specific needs.
 
-## Components
+## Getting Started
+
+### UDS
 
-### Getting Started
+UDS is the preferrred method for a full local deployment of LeapfrogAI. Instructions can be found on the [LeapfrogAI Documentation Site](https://docs.leapfrog.ai/docs/).
```
justinthelaw, line 40: Might be a good idea to mention that this is a Kubernetes deployment, even if our linked docs website mentions it later on.

jalling97: Updated!

README.md
```diff
@@ -6,14 +6,15 @@
 - [Table of Contents](#table-of-contents)
 - [Overview](#overview)
 - [Why Host Your Own LLM?](#why-host-your-own-llm)
+- [Getting Started](#getting-started)
+  - [UDS](#uds)
+  - [Tadpole](#tadpole)
```
YrrepNoj, line 11: We should remove the tadpole subsection, Tadpole is something we won't be supporting for a while.

README.md
```diff
@@ -32,34 +33,32 @@ Large Language Models (LLMs) are a powerful resource for AI-driven decision maki
...
+
+### API
 
 LeapfrogAI provides an API that closely matches that of OpenAI's. This feature allows tools that have been built with OpenAI/ChatGPT to function seamlessly with a LeapfrogAI backend.
 
 ### Backends
 
-> GitHub Repos:
-> | Repo | AMD64 Support | ARM64 Support | Cuda Support | Docker Ready | K8s Ready | Zarf Ready |
+> Available Backends:
```
YrrepNoj, line 54: We should update these to link to their place within the monorepo

website/content/en/docs/local deploy guide/quick_start.md
```diff
@@ -0,0 +1,86 @@
...
+weight: 2
+---
+
+# LeapfrogAI UDS Deployment
+
+The fastest and easiest way to get started with a deployment of LeapfrogAI is by using [UDS](https://github.com/defenseunicorns/uds-core). These quick start instructions show how to deploy LeapfrogAI in either a CPU or GPU-enabled environment.
+
+## Prerequisites
+
+- [K3D](https://k3d.io/)
```
YrrepNoj, line 13: k3d implies Docker but their instructions don't have you install docker. we can list `Docker` here to be explicit.

I wouldn't hate it if we didn't though.

website/content/en/docs/local deploy guide/quick_start.md
```diff
@@ -0,0 +1,86 @@
...
+
+GPU considerations (NVIDIA GPUs only):
+
+- NVIDIA GPU must have the most up-to-date drivers installed.
+- NVIDIA GPU drivers compatible with CUDA (>=12.2).
+- NVIDIA Container Toolkit is available via internet access, pre-installed, or on a mirrored package repository in the air gap.
+
+## Disclaimers
+
+It is recommended to run UDS as root.
```
YrrepNoj, line 24: Hmm? I don't think this is true?

website/content/en/docs/local deploy guide/quick_start.md
```diff
@@ -0,0 +1,86 @@
...
+
+GPU workloads **_WILL NOT_** run if GPU resources are unavailable to the pod(s). You must provide sufficient NVIDIA GPU scheduling or else the pod(s) will go into a crash loop.
+
+`whisper` can run without without GPU scheduling - just turn off the `GPU_ENABLED` boolean.
```
YrrepNoj, line 28:
```suggestion
`whisper` can run without GPU scheduling - just set the `GPU_LIMIT` value to `0`.
```

website/content/en/docs/local deploy guide/quick_start.md
YrrepNoj: As far as I know, the only time root is recommended (required) is when you are installing a `k3s` cluster when running the `zarf init` command.

website/content/en/docs/local deploy guide/quick_start.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+### CPU
+
+From within the cloned repository, deploy K3D and the LeapfrogAI bundle:
+
+``` bash
+cd bundles/cpu/
+uds create .
+uds deploy k3d-core-istio-dev:0.14.1
```
YrrepNoj, line 54: Maybe add a comment to check if a newer version exists.

YrrepNoj (changes requested):

website/content/en/docs/local deploy guide/quick_start.md
gphorvath: yeah we probably should list it

CollectiveUnicorn (dismissed):

jalling97:
As a last minute edit, I removed references to Tadpole from the deploy instructions in the docs and moved the Tadpole specific instructions to a draft so that it won't render on the site. This way if we do support Tadpole again the instructions will still be there but not actually visible. Let me know if that's a sufficient approach for the time being.

CollectiveUnicorn (approved):

website/content/en/docs/local deploy guide/quick_start.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+### CPU
+
+From within the cloned repository, deploy K3D and the LeapfrogAI bundle:
+
+``` bash
+cd bundles/cpu/
+uds create .
+uds deploy k3d-core-istio-dev:0.14.1      # be sure to check if a newer version exists
```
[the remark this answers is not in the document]
YrrepNoj: This is a public image so that shouldn't be a hard requirement. We could always suggest it as a best practice still.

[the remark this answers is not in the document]
justinthelaw: Hm, I actually tried it out and I wasn't able to pull it without logging in with a valid PAT.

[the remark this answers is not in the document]
jalling97: I personally haven't run into that issue, but I can add some text. I'm assuming the PAT needs at least `read:packages` access?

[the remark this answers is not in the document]
justinthelaw: Hm, I realized my PAT had expired on Friday and I had to make a new PAT and re-login via CLI in order to pull packages from GHCR. It is possible an expired PAT will cause a 403 error purely because it tries to authorize if a PAT is available in cred store. I think this is fine without the PAT!

justinthelaw (approved):

YrrepNoj (approved):

PR #331, opened 2024-03-29
vLLM - Swaps environment variable for confz .yaml configs and moves to src

CollectiveUnicorn:
- Replaces .env with new config.yaml configs.
- Moves all files into the src directory.
- Rebuilds requirements files
- Moves env/config copying out of the docker build process in favor of ARGs with defaults


CollectiveUnicorn:
Ruff fixes pending: https://github.com/defenseunicorns/leapfrogai/pull/333

packages/vllm/src/main.py
```diff
@@ -67,7 +83,66 @@ def remove_iterator(self, async_iterable):
             pass  # If the iterable is not found, ignore the error
 
 
-@LLM
```
YrrepNoj, line 70: I'm curious why we removed this decorator?

packages/vllm/Dockerfile
```diff
@@ -15,20 +29,24 @@ WORKDIR /home/leapfrogai
...
-    && apt-get -y install software-properties-common \
+    && apt-get install -y software-properties-common \
+    && apt-get -y update \
     && add-apt-repository universe \
-    && add-apt-repository ppa:deadsnakes/ppa \
+    && add-apt-repository ppa:deadsnakes/ppa  \
     && apt-get -y update
 
+# get deps for vllm compilation, model download, and pyenv
+RUN apt-get -y install git python3-venv make build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm libncurses5-dev libncursesw5-dev tk-dev libffi-dev
```
YrrepNoj, line 39: Is there documentation somewhere that defines these as deps?

CollectiveUnicorn: For everything to the right of `python3-venv` the source would be this page: https://github.com/pyenv/pyenv/wiki#suggested-build-environment there may be some things that we can remove that were assumed to be general dependencies for building pyenv from source files.

git is there so we can clone the repo and python3-venv to create a venv. Though now that I'm thinking about it we can probably remove venv since its in a container already. Though that path reference might need to get replaced.

YrrepNoj: That makes sense. I made a new issue for tracking this a little closer.

Related to #367

packages/vllm/README.md
```diff
@@ -3,7 +3,68 @@
...
+
+You can optionally specify different models or quantization types using the following Docker build arguments:
+
+- `--build-arg HF_HUB_ENABLE_HF_TRANSFER="1"`: Enable or disable HuggingFace Hub transfer (default: 1)
+- `--build-arg REPO_ID="TheBloke/Synthia-7B-v2.0-GPTQ"`: HuggingFace repository ID for the model
+- `--build-arg REVISION="gptq-4bit-32g-actorder_True"`: Revision or commit hash for the model
+- `--build-arg QUANTIZATION="gptq"`: Quantization type (e.g., gptq, awq, or empty for un-quantized)
+- `--build-arg TENSOR_PARALLEL_SIZE="1"`: The number of gpus to spread the tensor processing across
+
+### Run Locally
```
jalling97, line 34: We should probably standardize where these local instructions live in the monorepo. For now I've been putting them in the `README.md` at the root of the project, but given that some of the instructions can be a little lengthy, I like the approach here of putting it in the README for each project.

I was thinking about opening a new issue to move all of the local dev instructions for each package to their own corresponding `README.md`. @YrrepNoj what do you think?

justinthelaw: I think the idea of linking them at the root is a good idea! The instructions can live in each individual directoy with the root being a table of contents.

packages/vllm/src/main.py
YrrepNoj: I understand that something 'broke' and removing this decorator fixed it, but I am not comfortable merging this PR in until we understand why/what actually broke.

gphorvath (approved):

CollectiveUnicorn:
Issue found here to track the LLM issue: https://github.com/defenseunicorns/leapfrogai/issues/431

PR #335, opened 2024-04-02
ADR: Database ADR (0003)

gphorvath:


gerred (approved):

gerred (approved):

.github/workflows/e2e.yaml
```diff
@@ -12,6 +12,7 @@ on:
       # Ignore docs and website things
       - "!**.md"
       - "!docs/**"
+      - "!adr/**"
```
YrrepNoj, line 15: The `.github/workflow/e2e-shim.yaml` needs to be updated with the inverse of this path.

YrrepNoj (changes requested):

.github/workflows/e2e.yaml
gphorvath: fixed

gerred (dismissed):

PR #349, opened 2024-04-03
chore: bring ui into monorepo

### LINKED ISSUE #307: chore(ui): validate user input to forms and apis
# Validate user input to forms and apis

**Type:** Chore

**Description:** When a user submits a form or data to Svelte API endpoint, validate the input. Recommend using Yup or Zod.

**User Story:**
As a Leapfrog dev team member, I want my app to be protected from incorrect or malicious user input so that I can protect our mission heroes.

**Acceptance Criteria:**
- [ ] All user input to forms and API endpoints is validated
- [ ] Errors are properly handled for incorrect input
- [ ] Unit tests

### LINKED ISSUE #344: chore(ui): pull existing ui into monorepo
Until the new ui is ready, pull the old ui into the monorepo

andrewrisse:
Integrates the "new UI (lfaiui) into the monorepo.

New naming convention is "leapfrogai-ui" 

Includes Keycloak code that will not work until Supabase is deployed in our clusters.

We will leave leapfrog-demo deployed as is since it does not use Keycloak yet.

closes #344
closes #307


packages/ui/chart/templates/ui/deployment.yaml
```diff
@@ -0,0 +1,45 @@
...
+            - name: LEAPFROGAI_API_KEY
+              value: '###ZARF_VAR_LEAPFROGAI_API_KEY###'
+            - name: PUBLIC_SUPABASE_URL
+              value: '###ZARF_VAR_SUPABASE_URL###'
+            - name: PUBLIC_SUPABASE_ANON_KEY
+              value: '###ZARF_VAR_SUPABASE_ANON_KEY###'
+
+          resources:
+            requests:
+              memory: '5Gi'
```
gerred, line 41: 5gb RAM req/limits seems high esp. with a 0.5/1 vCPU req for the UI, has this been benchmarked? Don't want resource exhaustion if it's unnecessary.

packages/ui/chart/templates/ui/service.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: v1
+kind: ServiceAccount
```
gerred, line 2: SA should be unnecessary here

gerred (changes requested):
doing some initial change requests since this is gonna be a chonky review

src/leapfrogai_ui/package.json
```diff
@@ -0,0 +1,74 @@
...
+		"prettier-plugin-svelte": "^3.1.2",
+		"sass": "^1.71.1",
+		"svelte": "^4.2.12",
+		"svelte-check": "^3.6.0",
+		"tslib": "^2.4.1",
+		"typescript": "^5.4.2",
+		"vite": "^5.0.3",
+		"vitest": "^1.2.0"
+	},
+	"type": "module",
```
gerred, line 58: i might be out of date here, but does this still force the mjs extension on entrypoints? if it does I'm only seeing that in husky, does vite just work?

packages/ui/chart/templates/ui/deployment.yaml
andrewrisse: This was a copy paste from the previous UI, agree it seems high. I'll lower it, but no it has not been benchmarked (at least with the new UI). How about 1Gi/2Gi?

packages/ui/chart/templates/ui/service.yaml
andrewrisse: Copy paste from previous UI, removing.

src/leapfrogai_ui/package.json
andrewrisse: btw husky no longer works now that this is moved into the monorepo, we may want to create a new story to integrate at the monorepo version (if we want husky as a team): https://dev.to/mimafogeus2/enforce-git-hooks-in-monorepos-with-husky-but-how-3fma

I was using it for Trufflehog checks, linting/prettier, tests, and commitlint.

packages/ui/chart/templates/ui/deployment.yaml
gphorvath: should we just set to 0 for now?

andrewrisse:
> If Supabase is a resource we are sharing between the API DB and UI, do we need to move the Supabase directory up a level? Not totally sure how that works. @andrewrisse

That directory is for running Supabase locally with the UI although it does contain some migrations the frontend needs when setting up Supabase. I wonder if these could be moved up to a parent directory.

andrewrisse:
also closes #307

packages/ui/chart/templates/ui/deployment.yaml
andrewrisse: why 0? wouldn't it not run then?

gphorvath: I think 0 means no limit

https://kubesphere.io/blogs/understand-requests-and-limits-in-kubernetes/#:~:text=If%20it%20is%20set%20to,that%20of%20limits%20by%20default.

andrewrisse: I like to live dangerously

gerred:
> > If Supabase is a resource we are sharing between the API DB and UI, do we need to move the Supabase directory up a level? Not totally sure how that works. @andrewrisse
> 
> That directory is for running Supabase locally with the UI although it does contain some migrations the frontend needs when setting up Supabase. I wonder if these could be moved up to a parent directory.
> 
> @gerred I'm thinking consolating the supabase migrations and setup should be a separate issue once we have a sucessful deployment. I still need this in my section of the repo until that's in place

Agreed on migrations later.

packages/ui/chart/templates/ui/deployment.yaml
andrewrisse: @gerred do you have an opinion on the limits here?

gerred: This is really just eligibility for limits killing w/ pods - I'm fine with setting it to 0 until we have some Prometheus metrics to help dial it in.

src/leapfrogai_ui/.github/workflows/ISSUE_TEMPLATE/bug.md
YrrepNoj, line 1: Here and else-where: These files should be moved to the root `.github` or removed completely.

src/leapfrogai_ui/.husky/commit-msg
```diff
@@ -0,0 +1 @@
+npx --no -- commitlint --edit $1
```
YrrepNoj, line 1: Could the husky pre-commit things be merged into the existing `.pre-commit-config.yaml`?

andrewrisse: I can make that a separate story and attempt

src/leapfrogai_ui/.github/workflows/ISSUE_TEMPLATE/bug.md
andrewrisse: deleted for now, we should have a separate discussion to decide how we want to integrate a release pipeline.

gerred (approved):

gerred:
Read through all the Svelte parts, looks like a really good start. We should consider something a little more robust for migrations.

src/leapfrogai_ui/.husky/pre-commit
```diff
@@ -0,0 +1,10 @@
+docker run -v "$(pwd):/workdir" -i --rm trufflesecurity/trufflehog:latest git file:///workdir --only-verified --fail
+
+npm run format 
+npm run lint
+
+git add .
+
+if [ "$SKIP_TEST" != "true" ]; then
+    npm run test
```
YrrepNoj, line 9: Is this something we can create a workflow for that triggers on PRs?

andrewrisse: Removed husky, will re-attack pre-commit checks for monorepo in separate story

gerred (approved):

PR #361, opened 2024-04-05
Refactor project structure: multiple python projects in monorepo

jalling97:
- Breaks up the `pyproject.toml` at the root of the monorepo into multiple pyprojects. This way, each component of LFAI can be more easily built as its own package. It also minimizes the number of dependencies needed per package which should make them a little bit lighter weight.
- Decouples the API and SDK to minimize dependency sizes

A potential follow-on to this line of effort would be to determine a new project structure that takes most advantage of having multiple python projects in a monorepo.


gerred:
Took a quick early review, some new stuff to merge in from main for you, generally looking good.

packages/vllm/pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[project]
+name = "lfai-vllm"
+version = "0.1.0"
```
justinthelaw, line 3: Just want to make sure - should this be `0.6.1`? How are we doing versioning for these continuing packages as compared ot our latest `0.6.1` release? @jalling97 @YrrepNoj

packages/whisper/Dockerfile
```diff
@@ -13,9 +13,9 @@ RUN pip install ctranslate2
 RUN ct2-transformers-converter --model ${MODEL_NAME} --output_dir .model --copy_files tokenizer.json --quantization float32
```
justinthelaw, line 13: On line 11, above this segment, we install `transformers[torch]` and `ctranslate2` from outside of a pinned (with hashes generated) requirements file. I think we should keep these `dev` requirements within the `pyproject.toml` and use them here too. Both for consistency in case of breaking changes and slightly better supply chain control. This comment applies to any other Dockerfiles where we do this.

**Reference:**
```
9   # download and covnert OpenAI's whisper base
10  ARG MODEL_NAME=openai/whisper-base
11  RUN pip install transformers[torch]
12  RUN pip install ctranslate2
```

src/leapfrogai_api/pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[project]
+name = "leapfrogai-api"
+version = "0.6.0"
```
justinthelaw, line 3: Update version.

justinthelaw (changes requested):

packages/vllm/pyproject.toml
jalling97: Good point. I believe @YrrepNoj did tell me that everything should be bumped to `0.6.1`, so I'll set up dynamic versioning so they can all be the same.

packages/whisper/Dockerfile
jalling97: So what I ended up doing was pinning versions for `transformers[torch]` and `ctranslate2` in the whisper `pyproject.toml` and within the  `Dockerfile` pull those versions in and build the wheels locally. Instead of installing the model download dependencies outside of the venv, I just uninstalled them after downloading the model and before installing the `lfai-whisper` wheel. This is because I was having trouble with the pathing. It gets us the same outcome albeit slightly slower (by a few seconds, so I wasn't too concerned about it).

src/leapfrogai_sdk/chat/__init__.py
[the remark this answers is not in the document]
jalling97: As we discussed earlier, since we make the switch to an `src` structure, we technically should have needed them in order to be able to properly handle imports in the API. I'm still trying to figure out why it has been working despite the few missing `__init__.py` files. It may have something to do with the API being treated like a namespace, but I'm still figuring that one out. Ultimately, if we're installing each component as a standard Python package, then these `__init__.py` should be present.

justinthelaw (approved):

gphorvath (approved):

gerred (approved):

PR #366, opened 2024-04-05
API - Fixed typing for ChatFunction

jmho:
The current Pydantic typing for `parameters` in ChatFunction is set to type `Dict[str, str]` which is incorrect. This causes valid calls to the `/chat/completions` endpoint to fail if the request contains `parameters` that are not strings which is incorrect behavior. The following below is an example of a valid curl request made from OpenAI's sample request [here](https://platform.openai.com/docs/guides/function-calling?lang=python) that fails under the previous validation schema but passes after the changes on this PR are implemented.

Also, if you examine the [OpenAI Python API Library](https://github.com/openai/openai-python), you can see that the typing that [they](https://github.com/openai/openai-python/blob/0bdc3772a159f433a5c6c9bb18bfa3ea807599b4/src/openai/types/shared/function_parameters.py#L4) use for function parameters is `Dict[str, object]` so this should be updated to match.

```
curl -X 'POST' \
  'http://localhost:3000/openai/v1/chat/completions' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "model": "leapfrogai",
  "messages": [
    {
      "role": "string",
      "content": "string"
    }
  ],
  "functions": [
    {
      "name": "get_current_weather",
      "description": "Get the current weather in a given location",
      "parameters": {
        "type": "object",
        "properties": {
          "location": {
            "type": "string",
            "description": "The city and state, e.g. San Francisco, CA"
          },
          "unit": {
            "type": "string",
            "enum": ["celsius", "fahrenheit"]
          }
        },
        "required": ["location"]
      }
    }
  ],
  "temperature": 1,
  "stream": false,
  "stop": "string",
  "max_tokens": 128
}'
```


gerred:
Thanks @jmho! This is a good fix. To note, functions can be very dependent on functions tuning for models - OK with this PR since it's expanding the contract.

gerred (approved):

gerred:
@jmho We require [signed commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), can you update your commit with that?

jmho:
Just signed everything! Let me know if it all looks good now.

PR #376, opened 2024-04-09
fix: correct typos in e2e workflow path triggers

YrrepNoj:
Fixes a typo where the incorrect yaml shorthand was used for the `e2e.yaml` file.

Fixes a logical error when we ignored the wrong directory


gerred (approved):
nice catch

Issue #306, opened 2024-03-21
feat(ui): keycloak integration


andrewrisse:
# Keycloak Integration

**Type:** Feature

**Description:** Integrate Keycloak auth with Supabase

**User Story:**
As a mission hero using LeapfrogAI UI, I want to login with UDS Keycloak so that I can use the app as a logged in user.

**References:**
[Keycloak & Supabase](https://supabase.com/docs/guides/auth/social-login/auth-keycloak)

**Acceptance Criteria:**
- [ ] Keycloak auth is used with Supabase to authenticate users
- [ ] Deployed in test environment
- [ ] Configuration process is documented

YrrepNoj:
Added the blocked label. 

Blocked by #345

YrrepNoj:
completed; resolved #380

Issue #345, opened 2024-04-02
feat(backend): Create Supabase Zarf configuration


andrewrisse:
Deploy Supabase into leapfrog-04 and leapfrog-demo.

Given I am a Leapfrog team member, when I deploy LeapfrogAI systems I can connect them to Supabase in our clusters instead of a cloud SAAS offering.

- [x] Get the bitnami or community chart deployed locally
- [ ]  Determine what customizations (chart variables and UDS overrides) are needed for our use case
    - [ ] Auth
    - [ ] Ingress (istio) ->  uds-operator Package CRD
    - [ ] Pepr compliance
    - [ ] Note: potentially split this into a new issue for AWS provided backing store (s3) vs HA in-cluster backing store
- [x] Create Zarf package config that deploys said chart
- [ ] Update uds-bundle.yaml to include the supabase package
- [ ] Create new issues for work needed to utilize this package with our UI/Rag/etc

### Additional context
Official docs: https://supabase.com/docs/guides/self-hosting
Community Supported: https://supabase.com/docs/guides/self-hosting#community-supported
Reference for running with Docker: https://supabase.com/docs/guides/self-hosting/docker#running-supabase

CollectiveUnicorn:
Draft PR can be found here: https://github.com/defenseunicorns/leapfrogai/pull/380

YrrepNoj:
completed; resolved #380

PR #380, opened 2024-04-10
Add supabase deployment

CollectiveUnicorn:
Adds Supabase with initial configs to get it deployed via Zarf and UDS.
* Access the Supabase components via https://supabase-kong.uds.dev/
* Includes some basic e2e tests
* Readme with deployment instructions

Integrates Supabase with the UI using keycloaks SSO
* Automatically creates a default keycloak client with the id `uds-supabase` for integration with the UI

Closes
- https://github.com/defenseunicorns/leapfrogai/issues/345
- https://github.com/defenseunicorns/leapfrogai/issues/441
- https://github.com/defenseunicorns/leapfrogai/issues/448


packages/supabase/supabase-kubernetes/charts/supabase/README.md
```diff
@@ -0,0 +1,189 @@
+# Supabase for Kubernetes with Helm 3
```
YrrepNoj, line 1: @CollectiveUnicorn The [upstream chart](https://github.com/supabase-community/supabase-kubernetes/blob/main/charts/supabase/README.md) that this was taken from seems very customizable, is there any way for us to benefit from the team of devs maintaining the upstream chart and customize everything we need with the `values.yaml` and `uds override` features?

CollectiveUnicorn: Done

packages/supabase/zarf.yaml
```diff
@@ -0,0 +1,110 @@
...
+    supabase instance and all of its dependencies
+
+constants:
+  - name: IMAGE_VERSION
+    value: '###ZARF_PKG_TMPL_IMAGE_VERSION###'
+  - name: DASHBOARD_USERNAME
+    description: The dashboard's (Supabase studio) default username
+    value: supabase-admin
+
+variables:
```
justinthelaw, line 16: For the variables that start with `ENABLE_`, are those picked up automatically somewhere? I don't see `###ZARF_VAR_` anywhere in the values file, especially the storage section for persistence volumes and volume claims.

packages/supabase/manifests/declarative-conf-configmap.yaml
```diff
@@ -0,0 +1,170 @@
...
+        keyauth_credentials:
+            - key: {{SUPABASE_SERVICE_KEY}}
+
+    ###
+    ### Dashboard credentials
+    ###
+    basicauth_credentials:
+      - consumer: DASHBOARD
+        username: ###ZARF_CONST_DASHBOARD_USERNAME###
+        password: {{SUPABASE_DASHBOARD_PASSWORD}}
```
justinthelaw, line 21: When do these `SUPABASE_` prefixed variables get created? At application time or do they get produced somewhere else for consumption here?

I see that the `DASHBOARD` has a Secret resource we can apply, but what about the others? This also sort of ties into my comment later on about the Zarf variables that don't map to any of the values keys. If we, for say, enable REST, wouldn't that mean a Secret would need to be generated?

justinthelaw (commented):
Just some initial questions for my own understanding - looks great so far!

packages/supabase/manifests/declarative-conf-configmap.yaml
CollectiveUnicorn: `###ZARF_CONST_DASHBOARD_USERNAME###` is created at creation, as its a const. `{{SUPABASE_DASHBOARD_PASSWORD}}` is randomly generated at deployment via the secret config `supabase-dashboard-secret.yaml` and picked up in `bitnami-values.yaml` under `kong`

packages/supabase/zarf.yaml
CollectiveUnicorn: They should be in `bitnami-values.yaml` for each component. They were added after the initial commit so maybe they weren't in there when you checked. But each component should have some variable of `ENABLE_` under their `enabled: ` property like `###ZARF_VAR_ENABLE_KONG###`.

CollectiveUnicorn: They aren't configurable in `bitnami-values-basic.yaml` because that's for bootstrapping particular pieces of Supabase that will break without this step.

justinthelaw (dismissed):

justinthelaw (approved):

andrewrisse:
closes #306

YrrepNoj (changes requested):

packages/supabase/zarf.yaml
```diff
@@ -0,0 +1,110 @@
...
+    default: "false"
+  - name: ENABLE_KONG
+    description: 'Should the kong component be enabled'
+    default: "true"
+  - name: ENABLE_POSTGRES
+    description: 'Should the postgres component be enabled'
+    default: "true"
+  - name: EXTERNAL_KEYCLOAK_CLIENT_ID
+    description: 'External keycloak client value'
+    default: "lfaiui"
```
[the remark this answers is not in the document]
CollectiveUnicorn: Need to change this to leapfrogai, all leapfrog apps will fall under this client id.

packages/supabase/manifests/declarative-conf-configmap.yaml
```diff
@@ -0,0 +1,170 @@
...
+      - username: service_role
+        keyauth_credentials:
+            - key: {{SUPABASE_SERVICE_KEY}}
+
+    ###
+    ### Dashboard credentials
+    ###
+    basicauth_credentials:
+      - consumer: DASHBOARD
+        username: ###ZARF_CONST_DASHBOARD_USERNAME###
```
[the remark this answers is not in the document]
CollectiveUnicorn: Need to move into values.

.github/workflows/e2e.yaml
```diff
@@ -70,6 +70,26 @@ jobs:
...
 
+          ##########
+          # Supabase
+          ##########
+        - name: Deploy Supabase
+          run: |
+            make build-supabase LOCAL_VERSION=e2e-test
+            docker image prune -af
+            uds zarf tools kubectl create namespace leapfrogai
+            uds zarf tools kubectl create secret generic supabase-keycloak-secret --from-literal=secret='YOUR_KEYCLOAK_CLIENT_SECRET' -n leapfrogai
```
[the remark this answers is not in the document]
CollectiveUnicorn: It is a secret that is supposed to already exist in keycloak at startup. This should not be a dependency on running tests, will remove.

packages/supabase/bitnami-values-basic.yaml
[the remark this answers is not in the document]
CollectiveUnicorn: It is a values file for the same chart. Its not the same as everything is disabled but the jwt token generation job and the service account creation that it depends on.

There's a PR to shrink this down further https://github.com/defenseunicorns/leapfrogai/issues/456 that goes into more detail into why this exists.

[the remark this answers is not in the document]
CollectiveUnicorn: Will try modifying this to use the main values file.

packages/supabase/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+Replace `YOUR_KEYCLOAK_CLIENT_SECRET` with the actual client secret from Keycloak.
+
+## Step 3: Configure Keycloak Client URLs
+
+In the Keycloak client settings, ensure that the following URLs are added to both the "Valid Redirect URIs" and "Web Origins" sections:
+
+- `http://supabase-kong.uds.dev/auth/v1/callback`
+- `https://lfaiui.uds.dev/`
```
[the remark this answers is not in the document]
CollectiveUnicorn: @andrewrisse Going to remove this from here, so it'll need to go into the UI

packages/supabase/zarf.yaml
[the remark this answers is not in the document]
CollectiveUnicorn: @andrewrisse this val will need to be used in the UI instead of `lfaiui`

packages/supabase/zarf.yaml
```diff
@@ -0,0 +1,110 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/defenseunicorns/zarf/main/zarf.schema.json
+kind: ZarfPackageConfig
+metadata:
+  name: supabase
+  version: '###ZARF_PKG_TMPL_IMAGE_VERSION###'
+  description: >
+    supabase instance and all of its dependencies
+
+constants:
+  - name: IMAGE_VERSION
```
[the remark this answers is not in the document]
CollectiveUnicorn: I have removed this

.github/workflows/e2e.yaml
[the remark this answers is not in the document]
CollectiveUnicorn: This has been removed.

packages/supabase/zarf.yaml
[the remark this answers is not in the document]
CollectiveUnicorn: This has been updated.

packages/supabase/README.md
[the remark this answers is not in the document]
CollectiveUnicorn: Removed references to UI will try to use more generic communication that is not specific to the UI but applicable to applications in general integrating with Supabase.

Makefile
```diff
@@ -37,6 +37,10 @@ sdk-wheel: ## build wheels for the leapfrogai_sdk package as a dependency for ot
 	-rm ${SDK_DEST}/*.whl
 	python -m pip wheel src/leapfrogai_sdk -w ${SDK_DEST}
 
+build-supabase:
+	## Build the Zarf package
+	uds zarf package create packages/supabase -o packages/supabase --registry-override=ghcr.io=localhost:5000 --insecure --set IMAGE_VERSION=${LOCAL_VERSION} --confirm
```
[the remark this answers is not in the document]
CollectiveUnicorn: Opted to still use the local registry and make use of the `initial-push.sh` for that purpose. @YrrepNoj let me know if this is an acceptable solution.

packages/supabase/scripts/clean-cluster.sh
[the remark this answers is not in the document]
CollectiveUnicorn: This has been deleted.

packages/supabase/scripts/initial-push.sh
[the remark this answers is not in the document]
CollectiveUnicorn: This is now use during the make `build-supabase` step

packages/supabase/manifests/declarative-conf-configmap.yaml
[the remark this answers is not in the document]
CollectiveUnicorn: The const is now used for env vars in the values file instead of in the configmap directly.

packages/supabase/uds-package.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: uds.dev/v1alpha1
+kind: Package
```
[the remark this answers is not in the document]
CollectiveUnicorn: Added templating and new values to `bitnami-values.yaml` under `Values.leapfrogai`

packages/supabase/bitnami-values.yaml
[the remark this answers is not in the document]
CollectiveUnicorn: It now contains only the things that we're modifying and overlays on top of the original values file.

Resolving this: https://github.com/defenseunicorns/leapfrogai/issues/456

packages/ui/chart/templates/ui/deployment.yaml
```diff
@@ -39,7 +39,7 @@ spec:
               {{ else }}
               valueFrom:
                 secretKeyRef:
-                  name: supabase-jwt
+                  name: supabase-bootstrap-jwt
```
YrrepNoj, line 42: Question; what changed for secretKeyRef to change?

uds-bundles/dev/cpu/uds-config.yaml
```diff
@@ -21,4 +21,11 @@ variables:
...
-    supabase_anon_key: ''
\ No newline at end of file
+    supabase_anon_key: ''
+
+  supabase:
+    leapfrogai:
+      package:
+        host: supabase-kong
+        name: supabase
+        component: kong
```
YrrepNoj, line 31: This config is to set Zarf package variables but these variables don't exist in the `supabase/zarf.yaml`.
This looks to be the pathing of a manifest, were you intending this to be an 'UDS override`?

NOTE: Same comment for the other `uds-config.yaml`

packages/supabase/bitnami-values.yaml
YrrepNoj, line 1: Are you still intending on slimming this values file down?

packages/supabase/chart/templates/supabase-dashboard-secret.yaml
```diff
@@ -0,0 +1,21 @@
+{{- $dashboardPassword := randAlphaNum 10 }} 
+{{- $existingSecret := (lookup "v1" "Secret" .Release.Namespace "supabase-dashboard-secret") }}
+{{- if .Values.leapfrogai.testing }}
```
YrrepNoj, line 3: Let's pair up on this and see if we can come up with a way to abuse Zarf to do this better 🤔

YrrepNoj (changes requested):
A few more comments & observations. I will pull this down and follow the README shortly so I may have a few more comments on that file later.

packages/ui/chart/templates/ui/deployment.yaml
CollectiveUnicorn: Prevents crashing in the UI due to the helm chart name change

uds-bundles/dev/cpu/uds-config.yaml
CollectiveUnicorn: These are no longer needed, they are in the `bitnami-values.yaml`

packages/supabase/bitnami-values.yaml
CollectiveUnicorn: Will look into doing this for the main bitnami-values file.

Makefile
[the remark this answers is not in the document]
CollectiveUnicorn: Removed this script, opted to pull from remote for the build

packages/supabase/scripts/initial-push.sh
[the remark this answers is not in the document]
CollectiveUnicorn: Deleted this script

Makefile
```diff
@@ -37,6 +37,13 @@ sdk-wheel: ## build wheels for the leapfrogai_sdk package as a dependency for ot
 	-rm ${SDK_DEST}/*.whl
 	python -m pip wheel src/leapfrogai_sdk -w ${SDK_DEST}
 
+build-supabase: local-registry
+	## Pull down images and push them to the local registry (Zarf is super slow if the image is only in the local daemon)
+	./packages/supabase/scripts/initial-push.sh
+
+	## Build the Zarf package
+	uds zarf package create packages/supabase -o packages/supabase --registry-override=ghcr.io=localhost:5000 --insecure --set IMAGE_VERSION=${LOCAL_VERSION} --set TESTING=${TESTING:"false"} --confirm
```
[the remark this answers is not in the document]
CollectiveUnicorn: I have removed this and the script

packages/supabase/scripts/initial-push.sh
```diff
@@ -0,0 +1,35 @@
+docker pull supabase/studio:20230127-6bfd87b
```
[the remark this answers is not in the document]
CollectiveUnicorn: With the local registry no longer being used, its just a dev script and I've removed it.

uds-bundles/dev/cpu/uds-config.yaml
CollectiveUnicorn: These have been removed

packages/supabase/bitnami-values-basic.yaml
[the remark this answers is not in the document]
CollectiveUnicorn: That worked, using this as an overlay

packages/supabase/chart/templates/supabase-dashboard-secret.yaml
CollectiveUnicorn: Replace this with an env variable that is set before the test is run. The env variable contains the testing password by pulling it out of the cluster. The python test will then use that environment variable to do the health check.

packages/supabase/bitnami-values.yaml
CollectiveUnicorn: I've slimmed this one down as well.

.github/workflows/e2e.yaml
```diff
@@ -70,6 +70,31 @@ jobs:
...
+          run: |
+            make build-supabase LOCAL_VERSION=e2e-test
+            docker image prune -af
+            uds zarf tools kubectl create namespace leapfrogai
+            uds zarf package deploy packages/supabase/zarf-package-supabase-amd64-e2e-test.tar.zst --confirm
+            rm packages/supabase/zarf-package-supabase-amd64-e2e-test.tar.zst
+
+        - name: Test Supabase
+          run: |
+            DASHBOARD_PASSWORD=$(uds zarf tools kubectl get secret supabase-dashboard-secret -n leapfrogai -o jsonpath='{.data.password}' | base64 -d)
```
YrrepNoj, line 86: I think this is what you're trying to do?

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

CollectiveUnicorn: Ah yeah, I ended up doing it a little bit differently after realizing that what I had wasn't going to work in a single step.

CollectiveUnicorn: Updated it based on those docs.

packages/supabase/bitnami-values-bootstrap.yaml
```diff
@@ -0,0 +1,113 @@
...
+##
+
+## PostgreSQL chart configuration
+## ref: https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml
+## @param postgresql.enabled Switch to enable or disable the PostgreSQL helm chart
+## @param postgresql.auth.existingSecret Name of existing secret to use for PostgreSQL credentials
+## @param postgresql.architecture PostgreSQL architecture (`standalone` or `replication`)
+## @param postgresql.service.ports.postgresql PostgreSQL service port
+##
+postgresql:
```
justinthelaw, line 112: How does this value relate to the one in the `bitnami-values.yaml`, `line 176`? Are these bootstrap values merged or overwritten by the other values during the helm deployment?

packages/supabase/bitnami-values-bootstrap.yaml
```diff
@@ -0,0 +1,113 @@
+# Copyright VMware, Inc.
+# SPDX-License-Identifier: APACHE-2.0
+
+## @section Global parameters
+## Global Docker image parameters
+## Please, note that this will override the image parameters, including dependencies, configured to use the global value
+## Current available global Docker image parameters: imageRegistry, imagePullSecrets and storageClass
+##
+
+## @param global.imageRegistry Global Docker image registry
```
justinthelaw, line 10: Recommend using Zarf air-gap patterns using `imagePullSecrets.name: private-registry`, and (optional) also specifying each `image`'s repository with the `###ZARF_REGISTRY###` prefix. Without the pull secret,, the container will fail to pull images in an air-gap even with the mutating hook.

packages/ui/chart/templates/ui/deployment.yaml
YrrepNoj: I don't understand, how does this prevent crashing in the UI? Where did the helm charts name get changed?

packages/supabase/chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: uds.dev/v1alpha1
+kind: Package
+metadata:
+  name: {{ .Values.leapfrogai.package.name }}
+spec:
```
YrrepNoj, line 5: We should be using the `sso` key for this CRD.

packages/supabase/zarf.yaml
```diff
@@ -0,0 +1,112 @@
...
+      - docker.io/bitnami/kong:3.6.1-debian-12-r13
+  - name: supabase-post-process
+    description: "Perform necessary post processing here"
+    required: true
+    actions:
+      onDeploy:
+        before:
+          - cmd: kubectl delete cm supabase-kong-declarative-config -n leapfrogai 
+        after:
+          - cmd: kubectl rollout restart deployment supabase-kong -n leapfrogai
```
YrrepNoj, line 106: Restarting a deployment right after it gets deployed feels like a code smell. Why do we need to restart this deployment? Could this be solved by reordering how we deploy things?

YrrepNoj (changes requested):

packages/supabase/manifests/declarative-conf-configmap.yaml
```diff
@@ -0,0 +1,170 @@
...
+            config:
+              hide_credentials: true
+          - name: acl
+            config:
+              hide_groups_header: true
+              allow:
+                - admin
+                - anon
+
+      - name: realtime-v1
```
YrrepNoj, line 98: Our `packages/supabase/zarf.yaml` has a variable for `ENABLE_REALTIME`. It looks like this configmap is not templating itself depending on those variables.

Are there any potential issues we might run into if this configmap has this configuration for realtime when the person deploying sets `ENABLE_RELTIME` == false?

packages/supabase/zarf.yaml
```diff
@@ -0,0 +1,112 @@
...
+        version: 0.0.1
+        valuesFiles:
+          - "bitnami-values.yaml"
+      - name: supabase
+        namespace: leapfrogai
+        url: oci://registry-1.docker.io/bitnamicharts/supabase
+        releaseName: supabase
+        version: 4.0.1
+        valuesFiles:
+          - "bitnami-values.yaml"
```
YrrepNoj, line 85: This `bitnami-values.yaml` values file is being used by 3 different helm charts. This amount of overloading seems like it can get confusing and hard to maintain very quickly.

Can we please simplify this more?

packages/supabase/bitnami-values-bootstrap.yaml
justinthelaw: NVM! Only needed pre-init!

packages/supabase/chart/templates/uds-package.yaml
YrrepNoj: The Package CRD is [defined here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/crd/sources/package/v1alpha1.ts) and [here is an example of it being used](https://github.com/defenseunicorns/uds-core/blob/6e56b2afec8f54f8c0a4aa4b89fef1d1c754b627/src/grafana/chart/templates/uds-package.yaml#L7).

According to the documentation you have in the README of the PR and [this issue](https://github.com/defenseunicorns/leapfrogai/issues/448) I think the manifest should include something like this.

```yaml
spec:
  sso:
    - clientId: "supabase"
      name: "supabase"
      redirectUris:
        - "https://supabase-kong.uds.dev/auth/v1/callback"
        - "https://lfaiui.uds.dev/"
      webOrigins:
        - "https://supabase-kong.uds.dev"
        - "https://lfaiui.uds.dev"
```


NOTE: I followed the documentation in the issue you created but the `redirectUris` and `webOrigins` being the same lists feels a little wrong. Are we sure both of paths are needed in both those fields?

packages/supabase/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Step 1: Create a Client in Keycloak
+
+In the Keycloak management console, navigate to the "Clients" section and create a new client with a client ID of `leapfrogai`. Make sure to note down the client secret, as you'll need them later.
+
+## Step 2: Create a Kubernetes Secret
+
+Create a Kubernetes secret named `supabase-keycloak-secret` that contains the client secret from the previous step. Use the following command:
+
+```bash
```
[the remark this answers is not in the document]
CollectiveUnicorn: There's a plan to automate this in https://github.com/defenseunicorns/leapfrogai/issues/448 I know the secret is named `supabase-keycloak-secret` but most of the automation is focused around keycloak itself. Then this will be a minor addition as a crd.

packages/supabase/zarf.yaml
CollectiveUnicorn: We restart it because kong starts before the new declarative config is applied, meaning that it has the wrong configuration when it first starts. When the new configuration is applied kong must be restarted so that it can pick up the new config and load it into the container.

CollectiveUnicorn: Will add comments and todo for this.

packages/supabase/zarf.yaml
CollectiveUnicorn: There's a difference in how the chart deploys with helm vs Zarf/uds. Will create a ticket to investigate this further.

CollectiveUnicorn:
Will look into removing declarative config and requiring port forwarding to access the studio. Will need to update the test.

.github/workflows/e2e.yaml
```diff
@@ -70,6 +70,35 @@ jobs:
...
+
+        - name: Test Supabase
+          run: |
+            python -m pip install requests
+            python -m pytest ./tests/e2e/test_supabase.py -v
+
+        # This cleanup may need to be moved/removed when other packages depend on Supabase
+        - name: Cleanup Supabase
+          run: |
+            uds zarf package remove supabase -l=trace --confirm
```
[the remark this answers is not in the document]
CollectiveUnicorn: Created an issue for this here: https://github.com/defenseunicorns/leapfrogai/issues/479

packages/supabase/README.md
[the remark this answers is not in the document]
CollectiveUnicorn: This may be resolved even earlier by the UDS SSO configs that @YrrepNoj has suggested using

packages/supabase/bitnami-values-bootstrap.yaml
```diff
@@ -0,0 +1,113 @@
...
+  ## @param storage.enabled Enable Supabase storage
+  ##
+  enabled: false
+  ## Enable persistence using Persistent Volume Claims
+  ## ref: https://kubernetes.io/docs/concepts/storage/persistent-volumes/
+  ##
+  persistence:
+    ## @param storage.persistence.enabled Enable persistence using Persistent Volume Claims
+    ##
+    enabled: false
```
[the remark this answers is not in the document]
CollectiveUnicorn: This is just for the bootstrap of the jwt token, it creates a job, that job creates the secret and it ends so there's no need for persistence. The `bitnami-values.yaml` does have storage enabled and persistence by default. The default configs don't specify any storage-class in particular so it should be falling back to the default which is `local-path` I think atm. Here's the underlying values file just so you can see where this is happening: https://github.com/bitnami/charts/blob/8568db0ec0b217adb7ed53d615af3bfdb5de22f6/bitnami/supabase/values.yaml#L2458

packages/supabase/bitnami-values-bootstrap.yaml
CollectiveUnicorn: The base values file [here](https://github.com/bitnami/charts/blob/8568db0ec0b217adb7ed53d615af3bfdb5de22f6/bitnami/supabase/values.yaml#L2458) from the chart are applied first, the our changes in bitnami-values.yaml just so we can get resource configs and any other little changes that may also affect the jwt token job, then the `bitnami-values-bootstrap.yaml` is applied on top of that. This values file disables everything but the jwt token job so that it can be bootstrapped. This has to occur due to lifecycle issues. Namely that the original helm chart relies the jwt token job running during/before other parts of Supabase and when we do a Zarf deploy it does not occur until after everything has successfully started up.

packages/supabase/chart/templates/supabase-keycloak-secret.yaml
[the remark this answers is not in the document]
CollectiveUnicorn: Will do this and confirm that it works with everything else

packages/supabase/zarf.yaml
```diff
@@ -0,0 +1,112 @@
...
+  - name: ENABLE_VOLUME_PERMISSIONS
+    description: 'Should the volume permissions component be enabled'
+    default: "false"
+  - name: ENABLE_KONG
+    description: 'Should the kong component be enabled'
+    default: "true"
+  - name: ENABLE_POSTGRES
+    description: 'Should the postgres component be enabled'
+    default: "true"
+  - name: ENABLE_EXTERNAL_KEYCLOAK
```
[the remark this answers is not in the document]
CollectiveUnicorn: The word `external` is not used consistently in the upstream helm chart. There's no internal keycloak provided AFAIK, but there is an internal Postgres option. Its dependent on what part of Supabase we're talking about how `external` is interpreted.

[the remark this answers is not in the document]
CollectiveUnicorn: In this case it has more to do with gotrue then anything else. That's what kong is using under the covers I believe.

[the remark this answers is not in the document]
CollectiveUnicorn: https://github.com/netlify/gotrue

packages/supabase/zarf.yaml
```diff
@@ -0,0 +1,112 @@
...
+    description: 'Should the storage component be enabled'
+    default: "true"  
+  - name: ENABLE_STUDIO
+    description: 'Should the studio component be enabled'
+    default: "true"
+  - name: ENABLE_VOLUME_PERMISSIONS
+    description: 'Should the volume permissions component be enabled'
+    default: "false"
+  - name: ENABLE_KONG
+    description: 'Should the kong component be enabled'
```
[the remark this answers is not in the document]
CollectiveUnicorn: I'll refine the descriptions

packages/supabase/zarf.yaml
```diff
@@ -0,0 +1,112 @@
...
+      - docker.io/bitnami/supabase-storage:0.48.4-debian-12-r0
+      - docker.io/bitnami/supabase-studio:0.24.3-debian-12-r0
+      - docker.io/bitnami/kong:3.6.1-debian-12-r13
+  - name: supabase-post-process
+    description: "Perform necessary post processing here"
+    required: true
+    actions:
+      onDeploy:
+        before:
+          - cmd: kubectl delete cm supabase-kong-declarative-config -n leapfrogai
```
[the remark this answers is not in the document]
CollectiveUnicorn: Its a lifecycle issue, if we make `supabase-studio` not accessible by any means other than through port-forwarding then this may not be necessary.

packages/supabase/manifests/declarative-conf-configmap.yaml
CollectiveUnicorn: If you try to hit `http://supabase-kong.uds.dev/realtime/v1/*` it'll just throw an error if its disabled.

packages/supabase/chart/templates/uds-package.yaml
CollectiveUnicorn: Will validate the redirectUris and webOrigins are both needed once an initial working config is setup. If some of these are not necessary I'll strip them out.

packages/supabase/zarf.yaml
CollectiveUnicorn: Not sure if there's more that you'd potentially want with this. But the `bitnami-values.yaml` and `bitnami-values-bootstrap.yaml` files have been streamlined about as much as they can be. I could probably duplicate some things from bitnami-values to bitnami-values-bootstrap if we want to only have one values file but other then that its about as small as its going to get atm.

packages/supabase/zarf.yaml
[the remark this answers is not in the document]
CollectiveUnicorn: The descriptions should now better represent the individual pieces being enabled and disabled

packages/supabase/zarf.yaml
CollectiveUnicorn: Added comments, will look into removing if we can get the studio accessible through some other means

packages/ui/chart/templates/ui/deployment.yaml
CollectiveUnicorn: The name of the token is autogenerated from the chart name. The chart is named `supabase-bootstrap` so the name no longer matches. If we leave it `supabase-jwt` it'll fail since the secret is no longer called that.

packages/supabase/zarf.yaml
CollectiveUnicorn: Removed all of this code in favor of port forwarding

packages/ui/chart/templates/ui/deployment.yaml
YrrepNoj: If the secret name is autogenerated from the chart name, could we template this line here?

tests/e2e/test_supabase.py
```diff
@@ -0,0 +1,22 @@
+import requests
+import os
+
+health_urls = {
+    "auth_health_url": "http://supabase-kong.uds.dev/auth/v1/health",
+    "rest_health_url": "http://supabase-kong.uds.dev/rest/v1/",
+    "storage_health_url": "http://supabase-kong.uds.dev/storage/v1/status",
+}
+anon_api_key = os.environ["API_KEY"]
```
YrrepNoj, line 9: nit; could we leave a comment that explains what this key is (the anon-key for supabase?) so that people can read this file and know how to run the test locally?

People can figure this out if they read the `.github/workflows/e2e.yaml` so I will not block the merge on this comment so feel empowered to auto-resolve this comment if you don't think it's worth it.

packages/supabase/chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,32 @@
...
+  sso:
+    - name: Supabase
+      description: Client for logging into Supabase
+      clientId: uds-supabase
```
YrrepNoj, line 9: The `zarf.yaml` has a `###ZARF_CONST_ EXTERNAL_KEYCLOAK_CLIENT_ID###` constant. 

We should make sure that const has field for it set in the `bitnami-values.yaml` and then use that values template here.

tests/e2e/test_supabase.py
CollectiveUnicorn: Will update this

packages/supabase/chart/templates/uds-package.yaml
CollectiveUnicorn: I am confused as to what this means? Its used in the values file for `GOTRUE_EXTERNAL_KEYCLOAK_CLIENT_ID`.

YrrepNoj: This file is hardcoding the clientId to uds-supabase. Someone can updating the `zarf.yaml.EXTERNAL_KEYCLOAK_CLIENT_ID` to be something different.

This would mean that the GOTRUE things that supabase directs to will be different than the client that Pepr created in Keycloak.

justinthelaw (approved):

packages/supabase/chart/templates/uds-package.yaml
CollectiveUnicorn: Got it, will add it to values and use it in both places.

packages/supabase/zarf.yaml
CollectiveUnicorn: Added this back after our conversation

YrrepNoj (approved):
LGTM. This was a big effort so appreciate you working on this @CollectiveUnicorn!

PR #397, opened 2024-04-15
bug: use correct config file path when loading all configs

### LINKED ISSUE #347: fix: API does not pick up GPU pods on cluster restart
Bug
The API does not pickup existing pods on cluster restart. vLLM, llamacpp, text-embeddings, etc.. leading to a restart of the API being necessary.

Expected
When a backend is running, the API will pick it up within a reasonable amount of time. This should include on cluster restart.

Reproduce
Restart a cluster with GPU backends. Once the cluster is back up and running, check the api's models endpoint to see if they are detected.

YrrepNoj:
Fixes #347


justinthelaw (dismissed):

justinthelaw (commented):
I think the pytests will need to be changed. Just curious as to why this changed? Seems like you are moving around the same info but just placing them in a different argument.

gerred (changes requested):
tests

YrrepNoj:
@gerred @justinthelaw 

PR updated to fix pytest

gerred (approved):
lgtm!

justinthelaw (approved):

PR #405, opened 2024-04-17
370 model add repeater model to the monorepo

jalling97:
- Moves the repeater model into the monorepo
- Provides instructions for running the repeater model locally to evaluate the API

To do:
- Write out a make target for the repeater model


jalling97:
PR replaced by: https://github.com/defenseunicorns/leapfrogai/pull/423

justinthelaw (approved):

PR #424, opened 2024-04-19
feat(api): assistants endpoint

### LINKED ISSUE #292: feat(api): Assistants endpoints
### User Story

**As a** LeapfrogAI Developer 
**I want** an OpenAI compliant Assistants API
**So that** I can use the API to manage assistants for various applications.

### Acceptance Criteria

Implement and test the following endpoints:
- [ ] Create Assistant
- [ ] List Assistants
- [ ] Retrieve Assistants
- [ ] Modify Assistant
- [ ] Delete Assistant

### Additional context
- https://platform.openai.com/docs/api-reference/assistants

gphorvath:
closes #292


src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,117 @@
...
+            object="assistant",
+            tools=validate_tools_typed_dict(request.tools),
+            tool_resources=ToolResources.model_validate(request.tool_resources),
+            temperature=request.temperature,
+            top_p=request.top_p,
+            metadata=request.metadata,
+            response_format=request.response_format,
+        )
+    except Exception as exc:
+        raise HTTPException(
```
gphorvath, line 42: return exception vs raise exception

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,117 @@
...
+        assistant = Assistant(
+            id="",  # Leave blank to have Postgres generate a UUID
+            created_at=int(time.time()),
+            name=request.name,
+            description=request.description,
+            instructions=request.instructions,
+            model=request.model,
+            object="assistant",
+            tools=validate_tools_typed_dict(request.tools),
+            tool_resources=ToolResources.model_validate(request.tool_resources),
```
gphorvath, line 35: Might not need to double validate

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,117 @@
...
-async def list_assistants() -> list[Assistant]:
+async def list_assistants(session: Session) -> List[Assistant] | None:
     """List all the assistants."""
-    # TODO: Implement this function
-    raise HTTPException(status_code=501, detail="Not implemented")
+    try:
+        crud_assistant = CRUDAssistant(model=Assistant)
+        return await crud_assistant.list(client=session)
+    except FileNotFoundError as exc:
+        raise HTTPException(status_code=404, detail="No assistants found") from exc
```
gphorvath, line 62: Error would be something other than "no assistants found"

src/leapfrogai_api/data/crud_assistant_object.py
```diff
@@ -0,0 +1,87 @@
...
+            .execute()
+        )
+
+        _, response = data
+
+        if response:
+            return self.model(**response[0])
+        return None
+
+    async def delete(self, client: AsyncClient, assistant_id: str) -> AssistantDeleted:
```
gphorvath, line 74: add check that this can't overwrite ID

tests/integration/api/test_assistants.py
```diff
@@ -0,0 +1,102 @@
...
+    assert modify_response.status_code == 200
+    assert Assistant.model_validate(
+        modify_response.json()
+    ), "Should return a Assistant."
+    assert modify_response.json()["model"] == "test1", "Should be modified."
+
+    get_modified_response = client.get(
+        f"/openai/v1/assistants/{create_response.json()['id']}"
+    )
+    assert get_modified_response.status_code == 200
```
gphorvath, line 71: use status code object type

tests/integration/api/test_assistants.py
```diff
@@ -0,0 +1,102 @@
...
+
+    modify_response = client.post(
+        f"/openai/v1/assistants/{create_response.json()['id']}",
+        json=request.model_dump(),
+    )
+    assert modify_response.status_code == 200
+    assert Assistant.model_validate(
+        modify_response.json()
+    ), "Should return a Assistant."
+    assert modify_response.json()["model"] == "test1", "Should be modified."
```
gphorvath, line 66: convert to variable

src/leapfrogai_api/routers/openai/files.py
```diff
@@ -43,25 +41,38 @@ async def upload_file(
...
+        raise HTTPException(
+            status_code=500, detail="Failed to store file in bucket"
+        ) from exc
 
     return file_object
 
 
 @router.get("")
-async def list_files(session: Session) -> list[FileObject] | None:
+async def list_files(session: Session) -> ListFilesResponse | None:
```
YrrepNoj, line 64: Learning question; 

The `ListFilesResponse` class seems pretty bare only really only holds a `data` member that is typed to be a `list[FileObject]`. What benefit do we gain from using that Class instead of just returning a `list[FileObject]`?

src/leapfrogai_api/routers/openai/files.py
```diff
@@ -43,25 +41,38 @@ async def upload_file(
...
-        return await crud_file.list(client=session)
-    except FileNotFoundError as exc:
-        raise HTTPException(status_code=404, detail="No file objects found") from exc
+    crud_file = CRUDFileObject(model=FileObject)
+    crud_response = await crud_file.list(client=session)
+
+    if crud_response is None:
+        return None
+
+    return {
```
YrrepNoj, line 72: If we are going to declare that this function returns a `ListFilesResponse` can we return an actual instance of the `ListFilesResponse` Class instead of returning a dict that matches its current fields?

src/leapfrogai_api/routers/openai/files.py
```diff
@@ -43,25 +41,38 @@ async def upload_file(
...
-    try:
-        crud_file = CRUDFileObject(model=FileObject)
-        return await crud_file.list(client=session)
-    except FileNotFoundError as exc:
-        raise HTTPException(status_code=404, detail="No file objects found") from exc
+    crud_file = CRUDFileObject(model=FileObject)
+    crud_response = await crud_file.list(client=session)
+
+    if crud_response is None:
+        return None
```
YrrepNoj, line 70: What are your thoughts on returning a `ListFilesResponse` where it's data member is set to None instead of returning None itself?

YrrepNoj (commented):
PR is looking like it's going in the right direction. Left a few comments on some thoughts/questions I had.

src/leapfrogai_api/Makefile
```diff
@@ -4,3 +4,6 @@ install:
 dev:
 	make install
 	python -m uvicorn main:app --port 3000 --reload
+
+test-integration:
+	cd ../../ && python -m pytest tests/integration/api
```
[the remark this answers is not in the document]
gphorvath: I do think that would make sense

src/leapfrogai_api/utils/openai_util.py
```diff
@@ -0,0 +1,40 @@
...
+        if tool_type not in tool_mapping:
+            raise ValueError(f"Unknown tool type: {tool_type}")
+
+        tool_class = tool_mapping[tool_type]
+        tool_instance = tool_class(**tool_data)
+
+    if tool_instance is None:
+        raise ValueError("No tools specified.")
+
+    if len(data) > 128:
```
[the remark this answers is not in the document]
gphorvath: This just follow the arbitrary limit OpenAI set: https://platform.openai.com/docs/api-reference/assistants/createAssistant#assistants-createassistant-tools

I wasn't sure where to set our limit so I stuck to OpenAI's - although I'm sure theirs accounts for the other possible tool types.

src/leapfrogai_api/routers/openai/files.py
gphorvath: OpenAI returns a typed dict in the form:

``` python
{
    object: str
    data: list[FileObject]
}
```

I originally tried just returning `list[FileObject]`, but the OpenAI client library throws errors on that.

src/leapfrogai_api/utils/openai_util.py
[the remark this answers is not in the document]
gphorvath: It is now a named variable

src/leapfrogai_api/utils/openai_util.py
```diff
@@ -0,0 +1,40 @@
...
+tool_mapping = {
+    "code_interpreter": CodeInterpreterTool,
+    "file_search": FileSearchTool,
+    "function_tool": FunctionTool,
+}
+
+
+def validate_tools_typed_dict(data: List[Dict]) -> List[AssistantTool]:
+    """Validate a tool typed dict."""
+    for tool_data in data:
```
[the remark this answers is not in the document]
gphorvath: Nope, good catch!

src/leapfrogai_api/utils/openai_util.py
```diff
@@ -0,0 +1,40 @@
...
+    CodeInterpreterTool,
+    FileSearchTool,
+    FunctionTool,
+    AssistantTool,
+)
+
+tool_mapping = {
+    "code_interpreter": CodeInterpreterTool,
+    "file_search": FileSearchTool,
+    "function_tool": FunctionTool,
```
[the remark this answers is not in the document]
gphorvath: I kept it intentionally inline with the OpenAI spec here and tried to catch it in the response: https://github.com/defenseunicorns/leapfrogai/blob/9c4d013fb8e4917d9131a910ac764f80749627f9/src/leapfrogai_api/backend/types.py#L253

src/leapfrogai_api/routers/openai/files.py
gphorvath: Yes

src/leapfrogai_api/utils/openai_util.py
```diff
@@ -0,0 +1,40 @@
...
+            raise ValueError("Tool type not specified.")
+
+        tool_type = tool_data["type"]
+        if tool_type not in tool_mapping:
+            raise ValueError(f"Unknown tool type: {tool_type}")
+
+        tool_class = tool_mapping[tool_type]
+        tool_instance = tool_class(**tool_data)
+
+    if tool_instance is None:
```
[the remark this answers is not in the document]
gphorvath: Fixed via other comments

src/leapfrogai_api/Makefile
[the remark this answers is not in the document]
CollectiveUnicorn: That's the way I've seen it in the past, have test folders that are co-located with the tested code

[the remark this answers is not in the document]
gphorvath: moving this into an issue so it can be a standalone PR

https://github.com/defenseunicorns/leapfrogai/issues/514

src/leapfrogai_api/routers/openai/files.py
gphorvath: Good idea, made the change.

src/leapfrogai_api/README.md
```diff
@@ -23,4 +23,21 @@ Setup environment variables:
...
+## Integration Tests
+
+The integration tests serve to identify any mismatches between components:
+
+- Check all API routes
+- Validate Request/Response types
+- DB CRUD operations
+- Schema mismatches
+
+Integration tests require a Supbase instance and environment variables configured.
```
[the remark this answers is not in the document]
gphorvath: I made a reference to the Local Development section since that is how I'm doing it for now.

CollectiveUnicorn (changes requested):

tests/pytest/leapfrogai_api/test_api.py
```diff
@@ -74,6 +74,7 @@ def test_routes():
         "/openai/v1/embeddings": ["POST"],
         "/openai/v1/audio/transcriptions": ["POST"],
         "/openai/v1/files": ["POST"],
+        "/openai/v1/assistants": ["POST"],
```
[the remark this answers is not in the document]
gphorvath: The problem is it doesn't test "all" because of the need for unique keys in the dictionary. So it isn't actually giving us that certainty.

.gitignore
```diff
@@ -23,6 +23,7 @@ build/
 .ruff_cache
 .branches
 .temp
+src/leapfrogai_api/config.yaml
```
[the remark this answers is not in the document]
gphorvath: I'm happy to ignore them all, but I only knew with certainty that I wanted to ignore this one.

src/leapfrogai_api/data/crud_assistant_object.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __init__(self, model: type[Assistant]):
+        self.model = model
+
+    async def create(
+        self, client: AsyncClient, assistant: Assistant
+    ) -> Assistant | None:
+        """Create a new assistant."""
+        assistant_object_dict = assistant.model_dump()
+        if assistant_object_dict.get("id") == "":
```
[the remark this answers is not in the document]
gphorvath: Done

.gitignore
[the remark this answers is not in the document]
gphorvath: Done

src/leapfrogai_api/Makefile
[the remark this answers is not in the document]
alekst23: I am not sure if it's a good idea to mix the two. Normally you would have a `tests` directory at root level, and organize all testing there, something like this:
```
├── src/
├── tests/
│   ├── unit/
│   │   ├── ui/
│   │   ├── api/
│   │   ├── ...
│   ├── integration/
│   │   ├── ui/
│   │   ├── api/
│   │   ├── ...
│   ├── functional/
```

tests/pytest/leapfrogai_api/test_api.py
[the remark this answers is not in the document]
CollectiveUnicorn: Which endpoints are you referring to? For the assistants endpoint can't we add something like this?
- "/openai/v1/assistants": ["GET", "POST"]
- "/openai/v1/assistants/": ["GET", "POST", "DELETE"]

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,128 @@
...
+        assistant = Assistant(
+            id=assistant_id,
+            created_at=old_assistant.created_at,
+            name=request.name or old_assistant.name,
+            description=request.description or old_assistant.description,
+            instructions=request.instructions or old_assistant.instructions,
+            model=request.model or old_assistant.model,
+            object="assistant",
+            tools=validate_tools_typed_dict(request.tools) or old_assistant.tools,
+            tool_resources=ToolResources.model_validate(request.tool_resources)
```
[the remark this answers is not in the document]
CollectiveUnicorn: You're right, I missed that

tests/pytest/leapfrogai_api/test_api.py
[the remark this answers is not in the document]
gphorvath: I could rewrite the test a bit to support that, but it doesn't right now. Also it still wouldn't necessarily support something in the form `POST vector_stores/{vector_store_id}/files` (https://platform.openai.com/docs/api-reference/vector-stores-files/createFile)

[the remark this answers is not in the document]
gphorvath: We found a way

src/leapfrogai_api/data/crud_assistant_object.py
```diff
@@ -0,0 +1,89 @@
...
+            .execute()
+        )
+
+        _, response = data
+
+        if response:
+            return self.model(**response[0])
+        return None
+
+    async def list(self, client: AsyncClient) -> list[Assistant] | None:
```
andrewrisse, line 49: Should this only list assistants that belong to the user?
I think this might break once we add authentication unless our assistants table has "allow all" read permissions on it.
If we do only return a user's specific assistants, would this be a deviation from OpenAI? 

Maybe we should have our own endpoint that gets assistants by user. That's what the frontend currently does, I'm worried about performance issues if I have to pull all assistants in from the database, then filter to only that user's assistants

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,128 @@
...
+    session: Session, request: CreateAssistantRequest
+) -> Assistant:
     """Create an assistant."""
-    # TODO: Implement this function
-    raise HTTPException(status_code=501, detail="Not implemented")
+
+    try:
+        assistant = Assistant(
+            id="",  # Leave blank to have Postgres generate a UUID
+            created_at=int(time.time()),
```
andrewrisse, line 28: does supabase add the created_at field for us?

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,128 @@
...
+            tools=validate_tools_typed_dict(request.tools),
+            tool_resources=ToolResources.model_validate(request.tool_resources),
+            temperature=request.temperature,
+            top_p=request.top_p,
+            metadata=request.metadata,
+            response_format=request.response_format,
+        )
+    except Exception as exc:
+        raise HTTPException(
+            status_code=405, detail="Unable to parse assistant request"
```
andrewrisse, line 43: 405 is method not allowed (ex. GET vs POST), should this be a 400 maybe?

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,128 @@
...
+        ) from exc
+
+    try:
+        crud_assistant = CRUDAssistant(model=Assistant)
+        return await crud_assistant.create(assistant=assistant, client=session)
+    except Exception as exc:
+        raise HTTPException(
+            status_code=405, detail="Unable to create assistant"
```
andrewrisse, line 51: return a 500 instead of 405?

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,128 @@
...
-async def retrieve_assistant(assistant_id: str) -> Assistant:
+async def retrieve_assistant(session: Session, assistant_id: str) -> Assistant | None:
     """Retrieve an assistant."""
-    # TODO: Implement this function
-    raise HTTPException(status_code=501, detail="Not implemented")
+    try:
+        crud_assistant = CRUDAssistant(model=Assistant)
+        return await crud_assistant.get(assistant_id=assistant_id, client=session)
+    except FileNotFoundError as exc:
+        raise HTTPException(status_code=404, detail="Assistant not found") from exc
```
andrewrisse, line 77: Up for debate, but should we return a 200 or 204 with empty data when it's not found? 404 implies several possible conditions (ex. website down, proxy issues, etc). The request was successful but there was no data so 200 is what I've seen several backends do.

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,128 @@
...
-    assistant_id: str, request: ModifyAssistantRequest
+    session: Session, assistant_id: str, request: ModifyAssistantRequest
 ) -> Assistant:
     """Modify an assistant."""
-    # TODO: Implement this function
-    raise HTTPException(status_code=501, detail="Not implemented")
+
+    old_assistant = await retrieve_assistant(session, assistant_id)
+    if old_assistant is None:
+        raise HTTPException(status_code=404, detail="Assistant not found")
```
andrewrisse, line 88: same 404 comment as above, assumed for the rest of the 404s too

supabase/migrations/20240419164109_init-assistant.sql
```diff
@@ -0,0 +1,17 @@
...
+    metadata jsonb,
+    model text,
+    name text,
+    object text,
+    tools jsonb,
+    response_format text,
+    temperature float,
+    tool_resources jsonb,
+    top_p float
+  );
```
andrewrisse, line 17: Adding in my version of this for the frontend migrations for reference. Big differences are:
1. limiting the name, description, and model text lengths with VARCHAR to what openai limits them too
2. typed the created_at field differently
3.  "object" is type checked to be "assistant".
4. response_object can be a string or json. 
5. Specified not null fields.
```
create table assistants (
    id uuid primary key DEFAULT uuid_generate_v4(),
    object text check (object in ('assistant')),
    name varchar(255),
    description varchar(512),
    model varchar(255) not null,
    instructions TEXT,
    tools jsonb,
    tool_resources jsonb,
    metadata jsonb,
    temperature float,
    top_p float,
    response_format jsonb,
created_at timestamp with time zone default timezone('utc'::text, now()) not null
);
```

tests/integration/api/test_assistants.py
```diff
@@ -0,0 +1,113 @@
...
+        tools=[{"type": "file_search"}],
+        tool_resources={},
+        metadata={},
+        temperature=1.0,
+        top_p=1.0,
+        response_format="auto",
+    )
+
+    create_response = client.post("/openai/v1/assistants", json=request.model_dump())
+    assert create_response.status_code is status.HTTP_200_OK
```
andrewrisse, line 32: can we also use these status code constants in the main code (not just tests)?
status.HTTP_200_OK

andrewrisse (commented):
Final comment is just for error handling. Can we log out the errors so we can see them in the pods? I like modifying error responses returned to the caller for security reasons, but it's nice to see the actual error response in the logs where it's safe (ex. exactly what the supabase error was, not just 500 "Internal Error" or something.)

tests/integration/api/test_assistants.py
```diff
@@ -0,0 +1,113 @@
...
+from leapfrogai_api.backend.types import (
+    CreateAssistantRequest,
+    ModifyAssistantRequest,
+)
+
+client = TestClient(router)
+
+
+def test_assistants():
+    """Test creating an assistant. Requires a running Supabase instance."""
```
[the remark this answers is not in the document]
gphorvath: Broken up!

src/leapfrogai_api/routers/openai/assistants.py
gphorvath: Good point, I bet I could set a default via the schema, I'll have to check in the AM

tests/integration/api/test_files.py
```diff
@@ -0,0 +1,73 @@
...
+from fastapi import status
+from fastapi.testclient import TestClient
+from openai.types import FileObject, FileDeleted
+from leapfrogai_api.routers.openai.files import router
+
+client = TestClient(router)
+
+
+def test_files():
+    """Test creating a file. Requires a running Supabase instance."""
```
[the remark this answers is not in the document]
gphorvath: I modified this to be broken up similarly to the assistants tests.

src/leapfrogai_api/data/crud_assistant_object.py
gphorvath: @andrewrisse I added two tickets to cover this:

Authenticating user and returning relevant data: https://github.com/defenseunicorns/leapfrogai/issues/483

Pagination of OpenAI endpoints: https://github.com/defenseunicorns/leapfrogai/issues/530

Let's discuss more at Backlog grooming, but think maybe this is a 0.8.0 requirement, not necessarily a blocker for this PR.

src/leapfrogai_api/routers/openai/files.py
```diff
@@ -43,25 +41,41 @@ async def upload_file(
...
         )
-
     except Exception as exc:
-        raise HTTPException(status_code=500, detail="Failed to store file") from exc
+        crud_file_object.delete(file_id=file_object.id, client=client)
+        raise HTTPException(
+            status_code=500, detail="Failed to store file in bucket"
+        ) from exc
 
     return file_object
```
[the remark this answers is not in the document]
gphorvath: yes

src/leapfrogai_api/data/crud_assistant_object.py
```diff
@@ -0,0 +1,89 @@
...
+
+        if response:
+            return [self.model(**item) for item in response]
+        return None
+
+    async def update(
+        self, client: AsyncClient, assistant_id: str, assistant: Assistant
+    ) -> Assistant | None:
+        """Update an assistant by its ID."""
+        data, _count = (
```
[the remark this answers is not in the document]
gphorvath: This is handled by the Modify Assistant route. I added some detail there to specify what can be updated.

CollectiveUnicorn (dismissed):

CollectiveUnicorn (approved):

CollectiveUnicorn (dismissed):

src/leapfrogai_api/utils/openai_util.py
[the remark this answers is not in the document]
alekst23: We should add a unit test to make sure this value is validated

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,157 @@
...
+    try:
+        assistant = Assistant(
+            id="",  # This is set by the database to prevent conflicts
+            created_at=0,  # This is set by the database
+            name=request.name,
+            description=request.description,
+            instructions=request.instructions,
+            model=request.model,
+            object="assistant",
+            tools=validate_tools_typed_dict(request.tools),
```
justinthelaw, line 33: It seems like the OpenAI Python library does have a set of Pydantic classes built-in as sub-fields for the Assistant Pydantic model. If you use the Assistant pydantic model from the 1.21.1 OpenAI Python lib, you don't need to build a custom validator.

Assistant model: https://github.com/openai/openai-python/blob/bb95fe35ad4b480ebce57aa5b48df620df8d761b/src/openai/types/beta/assistant.py#L78

AssistantTool model: https://github.com/openai/openai-python/blob/v1.21.1/src/openai/types/beta/assistant_tool.py

and

Example tool's model: https://github.com/openai/openai-python/blob/v1.21.1/src/openai/types/beta/code_interpreter_tool.py

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,157 @@
...
+            created_at=old_assistant.created_at,
+            name=request.name or old_assistant.name,
+            description=request.description or old_assistant.description,
+            instructions=request.instructions or old_assistant.instructions,
+            model=request.model or old_assistant.model,
+            object="assistant",
+            tools=validate_tools_typed_dict(request.tools) or old_assistant.tools,
+            tool_resources=ToolResources.model_validate(request.tool_resources)
+            or old_assistant.tool_resources,
+            temperature=request.temperature or old_assistant.temperature,
```
gphorvath, line 127: Edge case: If temperature = 0, it might inadvertently default to old_assistant.temperature

`request.get("temperature", old_assistant.temperature)`

src/leapfrogai_api/routers/openai/assistants.py
justinthelaw: An exmaple of when you may want to build your own Pydantic class (model) and want to build a validator for a particular field:

```python
class Input(BaseModel):
    input: str

    @validator("input")
    def max_token_length(cls, v):
        """
        The `max_token_length` function is a validator in Python that checks if the tokenized length of the
        input exceeds a maximum length and raises a `ValueError` if it does.

        :param cls: The `cls` parameter refers to the class that the `max_token_length` method is defined
        in. In this case, it is a class that contains the `max_token_length` method
        :param v: The parameter `v` represents the input value that is being validated. In this case, it is
        the `input` value that is being passed to the `max_token_length` function
        :return: The variable `v` is being returned.
        """
        tokenizer = T5Tokenizer.from_pretrained(
            TOKENIZER,
            legacy=False,
        )
        tokenized_length = len(tokenizer.tokenize(v))

        if tokenized_length > MAX_INPUT_TOKEN_LENGTH:
            max_length_error = f"The maximum token length is {MAX_INPUT_TOKEN_LENGTH}"
            raise ValueError(max_length_error)

        if tokenized_length < MIN_INPUT_TOKEN_LENGTH:
            min_length_error = f"The minimum token length is {MIN_INPUT_TOKEN_LENGTH}"
            raise ValueError(min_length_error)

        return v
```

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,157 @@
...
+            detail="Unable to parse assistant request",
+        ) from exc
+
+    try:
+        return await crud_assistant.update(
+            db=session, object_=new_assistant, id_=assistant_id
+        )
+    except FileNotFoundError as exc:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST, detail="Assistant not found"
```
gphorvath, line 144: Internal server error

src/leapfrogai_api/routers/openai/files.py
```diff
@@ -24,73 +21,71 @@ async def upload_file(
         file_object = FileObject(
             id="",  # This is set by the database to prevent conflicts
             bytes=request.file.size,
-            created_at=int(time.time()),
+            created_at=123,  # This is set by the database to prevent conflicts
```
gphorvath, line 24: 0

src/leapfrogai_api/routers/openai/assistants.py
gphorvath: found a way to do this!

src/leapfrogai_api/routers/openai/assistants.py
gphorvath: Fixed

src/leapfrogai_api/routers/openai/assistants.py
gphorvath: We worked together to kill this code and do it better

CollectiveUnicorn (approved):

justinthelaw (dismissed):
Was IRL witness, can confirm new tests/types work haha

andrewrisse (approved):

src/leapfrogai_api/routers/openai/files.py
```diff
@@ -24,73 +21,71 @@ async def upload_file(
...
+        crud_file_object.delete(db=session, id_=file_object.id)
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Failed to store file",
+        ) from exc
 
 
 @router.get("")
-async def list_files(session: Session) -> list[FileObject] | None:
+async def list_files(session: Session) -> ListFilesResponse:
```
alekst23, line 57: @gphorvath Why do we need to return an object here instead of a list? It seems to have one other member besides the `data` list, which is `object` . Are we envisioning something processing this list differently based on the `object` value that is returned? Seems like it's defaulting to `list`. How are we planning to use that value?

gphorvath: This is directly inherited from the OpenAI spec.

YrrepNoj (approved):

.gitignore
```diff
@@ -23,6 +23,7 @@ build/
 .ruff_cache
 .branches
 .temp
+config.yaml
```
alekst23, line 26: Are we trying to omit all `config.yaml` files in the project? We have some files like `src/leapfrogai_sdk/config.py`, `src/leapfrogai_api/utils/config.py`, and others which would be affected and no longer be tracked for changes. Are these local dev configs?

gphorvath: config.py should be fine, but I double checked and there are config.yaml's in some of packages, so I should probably specify just the API. I think I had it that way initially actually, then changed it.

gphorvath: I changed it back!

CollectiveUnicorn (dismissed):

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,194 @@
...
+                )
+
+    if request.tool_resources is not None:
+        for tool_resource in request.tool_resources:
+            if tool_resource is ToolResourcesCodeInterpreter:
+                if tool_resource["file_ids"] is not None:
+                    raise HTTPException(
+                        status_code=status.HTTP_400_BAD_REQUEST,
+                        detail="Code interpreter tool is not supported",
+                    )
```
alekst23, line 38: There are a few issues in this block. One is the use of `is` keyword, which is generally used for checking object instances, not for type checking. The other is `is not None`, which will fail for empty lists ([] != None)

We should use a pattern similar to this:
```python
    if request.tool_resources:
        if list(filter(lambda resource: isinstance(resource, ToolResourcesCodeInterpreter) and resource.get("file_ids"), request.tool_resources)):
            raise HTTPException(
                status_code=status.HTTP_400_BAD_REQUEST,
                detail="Code interpreter tool is not supported",
            )
```

This will properly handle type checking and truthiness validation for properties.

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,194 @@
...
-    # TODO: Implement this function
-    raise HTTPException(status_code=501, detail="Not implemented")
+
+    if request.tools is not None:
+        for tool in request.tools:
+            if tool.type not in ["file_search"]:
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail=f"Unsupported tool type: {tool.type}",
+                )
```
alekst23, line 29: Similar comment to the one below (going bottom up because why not)
```python
    SUPPORTED_TOOLS = ["file_search"]
    if request.tools and (unsupported_tool := next((tool for tool in request.tools if tool.type not in SUPPORTED_TOOLS), None)):
        raise HTTPException(
            status_code=status.HTTP_400_BAD_REQUEST,
            detail=f"Unsupported tool type: {unsupported_tool.type}",
        )
```

This will properly check empty values

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,194 @@
...
+
+    if request.tool_resources is not None:
+        for tool_resource in request.tool_resources:
+            if tool_resource is ToolResourcesCodeInterpreter:
+                if tool_resource["file_ids"] is not None:
```
alekst23, line 135: Same comment as above here. We should use `isinstance` and `get` functions

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,194 @@
...
+                    raise HTTPException(
+                        status_code=status.HTTP_400_BAD_REQUEST,
+                        detail="Code interpreter tool is not supported",
+                    )
+
+    crud_assistant = CRUDAssistant(model=Assistant)
+
+    old_assistant = await crud_assistant.get(db=session, id_=assistant_id)
+    if old_assistant is None:
```
alekst23, line 144: Might be safer to avoid `is None` like this:
```python
    if old_assistant := await crud_assistant.get(db=session, id_=assistant_id):
```

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -1,47 +1,201 @@
...
+            tools=request.tools or old_assistant.tools,
+            tool_resources=ToolResources.model_validate(request.tool_resources)
+            or old_assistant.tool_resources,
+            temperature=float(request.temperature)
+            if request.temperature is not None
+            else old_assistant.temperature,
+            top_p=request.top_p or old_assistant.top_p,
+            metadata=request.metadata or old_assistant.metadata,
+            response_format=request.response_format or old_assistant.response_format,
+        )
```
alekst23, line 174: We will have issues with this code in the following instances:
1. Empty property values in `request` will be ignored. So you will not be able to clear any existing values.
2. If the property is missing from the object, it will thrown an error

We should use the following pattern:
```python
new_assistant = Assistant(
    id=assistant_id,
    created_at=old_assistant.created_at,
    name=getattr(request, 'name', old_assistant.name),
    description=getattr(request, 'description', old_assistant.description),
    instructions=getattr(request, 'instructions', old_assistant.instructions),
    model=getattr(request, 'model', old_assistant.model),
    object="assistant",
    tools=getattr(request, 'tools', old_assistant.tools),
    tool_resources=ToolResources.model_validate(getattr(request, 'tool_resources', None)) or old_assistant.tool_resources,
    temperature=float(getattr(request, 'temperature', old_assistant.temperature)),
    top_p=getattr(request, 'top_p', old_assistant.top_p),
    metadata=getattr(request, 'metadata', old_assistant.metadata),
    response_format=getattr(request, 'response_format', old_assistant.response_format),
)

```

alekst23 (approved):
👍

CollectiveUnicorn (approved):

YrrepNoj (approved):

PR #436, opened 2024-04-24
fix: send and cancel btn tooltips

andrewrisse:
Carbon applies the iconDescription as a tooltip for buttons that use icons. This label is attached as an aria-label to the svg component within the button, not the actual button. We need a test id to get the buttons for tests, while also ensuring the tooltips are not undefined for the user. I didn't want to put an aria-label on the button as well as the svg, so this solution fixes the tooltips while also ensuring tests pass.


gerred (approved):

justinthelaw:
Recommend merging with base branch before merge, otherwise lgtm!!

gerred (approved):

PR #437, opened 2024-04-24
chore: refactoring API and adding stubs for new OpenAI endpoints

gphorvath:
Adds stubs for new assistants endpoint
Refactors API into separate routers for each route


gerred (changes requested):
One thing we should be consistent on is using `Union[x,y]` vs. `x | y` in type signatures. We had users pinned on Python 3.8 (maybe now less relevant with our direciton) before type syntax was built into Python. I'm fine with moving either direction, but this PR should unify on one or the other.

Other than that, lgtm.

CollectiveUnicorn (approved):

gerred (approved):
lgtm!

PR #439, opened 2024-04-24
407 Add test workflow for leapfrogai_ui

jalling97:
- Adds a workflow for running the existing unit tests within the new UI.
- Adds a workflow to scan for active secrets using trufflehog
- Adds UI-focused linting checks to the existing lint workflow
- Adds eslint and prettier to pre-commit checks

After the workflows were added, linting changes were applied to this PR to get the UI compliant with the linter/formatter


.github/workflows/secret-scan.yaml
```diff
@@ -0,0 +1,20 @@
+name: secret-scan
+on:
+  pull_request:
+    paths:
+      # Catch-all
+      - "**"
```
jalling97, line 6: Is there anytime we don't want to run this workflow? At the moment it runs on all PRs always.

andrewrisse:
I disabled the 'no-unde' rule for eslint because Typescript is already checking for this and gives no errors. 
Docs Reference: https://typescript-eslint.io/troubleshooting/#i-get-errors-from-the-no-undef-rule-about-global-variables-not-being-defined-even-though-there-are-no-typescript-errors

.github/workflows/lint.yaml
```diff
@@ -21,3 +21,14 @@ jobs:
         with:
           version: 0.3.4
           args: format --check
+
+      # Javascript & Typescript Linting
+      - uses: actions/setup-node@v4
```
justinthelaw, line 26: Need hash pinning instead

.github/workflows/lint.yaml
```diff
@@ -21,3 +21,14 @@ jobs:
...
+        with:
+          node-version: 18
```
justinthelaw, line 28: Recommend pinning using ASoT, e.g., `package.json` or other similar method.

.github/workflows/secret-scan.yaml
justinthelaw: I think it's good to always run this, because anything could be accidentally pasted or added.

.github/workflows/secret-scan.yaml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  secret-scan: # scan for any live secrets in the repository using trufflehog
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      with:
+        fetch-depth: 0
+    - name: Secret Scanning
+      uses: trufflesecurity/trufflehog@main
```
justinthelaw, line 18: Recommend pinning via hash.

.github/workflows/ui-test.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+jobs:
+  ui-test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - uses: actions/setup-node@v4
```
justinthelaw, line 27: Need hash pinning instead

.github/workflows/ui-test.yaml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          node-version: 18
```
justinthelaw, line 29: Same comment as before about node version pinning

.pre-commit-config.yaml
```diff
@@ -21,3 +21,18 @@ repos:
     hooks:
       - id: ruff # Run the linter.
       - id: ruff-format # Run the formatter.
+
+  # ESlint
+  - repo: https://github.com/pre-commit/mirrors-eslint
```
justinthelaw, line 26: Recommend added deps, not sure if ESLint pre-commit actually uses the local config...

```yaml
    - repo: https://github.com/pre-commit/mirrors-eslint
      rev: v8.56.0
      hooks:
          - id: eslint
            name: ESLint and Prettier
            files: \.([jt]sx?|svelte)$ # *.js, *.jsx, *.ts, *.tsx and *.svelte
            types: [file]
            additional_dependencies:
                - eslint@8.44.0
                - eslint-plugin-prettier@5.0.0
                - eslint-plugin-svelte@2.35.1
            args: [--fix]
```

justinthelaw (commented):
Just a question and some minor things!

gerred (approved):

.github/workflows/lint.yaml
```diff
@@ -21,3 +21,14 @@ jobs:
...
+
+      - name: Install UI Dependencies
+        run: npm --prefix src/leapfrogai_ui install
```
andrewrisse, line 31: I think we should be using ```npm ci``` instead of ```npm install``` for a workflow.
https://stackoverflow.com/a/53325242/11761227

.pre-commit-config.yaml
gerred: lgtm pending checking this

gerred:
all lgtm pending @justinthelaw's feedback being resolved.

.github/workflows/ui-test.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Install Dependencies
+        run: npm --prefix src/leapfrogai_ui install
```
andrewrisse, line 32: same comment as above about npm install vs npm ci

.github/workflows/lint.yaml
jalling97: Why hash pinning? Does it have to do with pointing us directly to a specific commit instead of a release tag that could be overridden?

.pre-commit-config.yaml
andrewrisse: there's also a prettierrc file we would want it to use. Can we confirm it does pull those?

.github/workflows/secret-scan.yaml
jalling97: Why for trufflehog? Wouldn't we want to always keep these kinds of scans as current as possible?

.github/workflows/lint.yaml
justinthelaw: Oh, I'm talking about the node setup action, not trufflehog

.github/workflows/lint.yaml
andrewrisse: ya and it uses the package-lock.json to do a "clean install", it's what I've always seen used in pipelines. If there issues with the dependencies (missing or mismatched) it will error out instead of installing. It will just result in a consistent build process, no surprises from cache issues

.github/workflows/secret-scan.yaml
gerred: Is trufflehog's scanning self contained into the container or is it calling out to an external service for secret scanning?

.github/workflows/lint.yaml
jalling97: I see the reason for pinning the `setup-node` version, I think my trufflehog question was in another comment thread

.github/workflows/secret-scan.yaml
gerred: disregard, my GH isn't updating new comments.

.pre-commit-config.yaml
```diff
@@ -21,3 +21,18 @@ repos:
...
+      - id: eslint
+        files: \.(js|jsx|ts|tsx|svelte)$ # apply eslint to the provided filetypes
+        types: [file]
+
+  # Prettier Code formatter
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: "v3.1.0"
+    hooks:
+      - id: prettier
+        files: \.(js|jsx|ts|tsx|svelte)$ # apply prettier to the provided filetypes
```
gerred, line 38: do we need .cjs and .mjs here?

.pre-commit-config.yaml
andrewrisse: replacing prettier with eslint-plugin-prettier resolves this conversation

jalling97: I also confirmed that the ESlint precommit hook does use our local `eslint` and `prettierrc` configs.

.pre-commit-config.yaml
jalling97: Added :+1:

.github/workflows/lint.yaml
jalling97: Pinned the node version in `package.json` and changed the checks to reference that version :+1:

.github/workflows/secret-scan.yaml
justinthelaw: @jalling97 I think I mistakenly added this comment to this job. I am good here! Seems like @gerred still has a question though.

jalling97: Unless I missed something obvious, I couldn't really find anything concrete as to whether or not the scanning reaches out to an external service, but I did test the scans without being connected to the internet and they still worked so maybe?

gerred:
fix conflicts then lgtm

gerred (approved):

gerred (dismissed):

.github/workflows/secret-scan.yaml
YrrepNoj, line 6: This doesn't hurt to keep, but it is a little confusing because it's overriding the default behavior with a catchall that does the same thing. 

Let's just remove this if we aren't going to actually be restricting on specific paths.

.github/workflows/ui-test.yaml
```diff
@@ -0,0 +1,35 @@
+name: ui-test
+on:
+  pull_request:
+    paths:
```
YrrepNoj, line 4: I would like to make this test be a blocker for PRs, if we have the `paths` we will need a 'shim' workflow.

YrrepNoj (changes requested):
Two small change requests.

.github/workflows/secret-scan.yaml
jalling97: I thought about taking it out for that reason, but part of me thought this was actually more clear about when and where the secret-scanner gets ran. Happy to remove it if that doesn't make sense/isn't worth it though.

.github/workflows/ui-test.yaml
jalling97: Good point. I can add a `ui-test-shim.yaml` workflow

.github/workflows/secret-scan.yaml
jalling97: Ended up removing it.

YrrepNoj (approved):
LGTM.

PR #446, opened 2024-04-29
chore: adding basic e2e-test for supabase

gphorvath:


.github/workflows/e2e.yaml
```diff
@@ -70,6 +70,26 @@ jobs:
...
+          run: |
+            make build-supabase LOCAL_VERSION=e2e-test
+            docker image prune -af
+            uds zarf tools kubectl create namespace leapfrogai
+            uds zarf tools kubectl create secret generic supabase-keycloak-secret --from-literal=secret='YOUR_KEYCLOAK_CLIENT_SECRET' -n leapfrogai
+            uds zarf package deploy packages/supabase/zarf-package-supabase-amd64-e2e-test.tar.zst --confirm
+            rm packages/supabase/zarf-package-supabase-amd64-e2e-test.tar.zst
+
+        # This cleanup may need to be moved/removed when other packages depend on Supabase
+        - name: Cleanup Supabase
```
YrrepNoj, line 86: Is there anything we can check before we remove supabase? 

Super great to at least test that we can build & deploy supabase, but it would be even better if there was a healthcheck or some simple service that we could query to take just one step further in our validation.

CollectiveUnicorn: That could be a separate PR, I could add it to the list of next Supabase work.

CollectiveUnicorn (approved):

.github/workflows/e2e.yaml
YrrepNoj: @gphorvath why not check against the health endpoint in this PR? That seems like a small step that provides a large amount of value.

gphorvath: In theory that is happening as a part of the deployment, but I can test again

PR #464, opened 2024-04-30
408 ci update publish workflow w leapfrogai UI artifacts

jalling97:
- Adds the UI zarf build process to the release workflow
- Updates the Makefile with new ui target
- Updates UI zarf package yamls


Makefile
```diff
@@ -53,6 +53,18 @@ build-api: local-registry setup-api-deps ## Build the leapfrogai_api container a
 	uds zarf package create packages/api -o packages/api --registry-override=ghcr.io=localhost:5000 --insecure --set LEAPFROGAI_IMAGE_VERSION=${LOCAL_VERSION} --confirm
 
 
+build-ui: ## Build the leapfrogai_ui container and Zarf package
```
YrrepNoj, line 56: With this new make target, I tthink  it make sense to remove the Makefile at `src/leapfrogai_ui/Makefile`.

YrrepNoj (changes requested):
Thanks for the PR. Just one single change request!

Makefile
jalling97: Good catch. I'll remove the old one and update any documentation referencing it

jalling97: Removed :+1:

YrrepNoj (approved):

PR #466, opened 2024-05-01
chore: reformat with new prettier settings

andrewrisse:
Updated to not use tabs because I'm not a monster.

Also changed tabWidth to 2.


justinthelaw (approved):
Ah yes, the best prettier settings!

gerred (approved):

Issue #422, opened 2024-04-18
feat(api): Vector Store Files endpoints


gphorvath:
### User Story

**As a** LeapfrogAI Developer 
**I want** an OpenAI compliant Vector Stores Files API
**So that** I can add, remove, and list files in the vector store

### Acceptance Criteria

Implement and test the following endpoints:
- [ ] Create vector store file
- [ ] List vector store files
- [ ] Delete vector store files

### Additional context
- https://platform.openai.com/docs/api-reference/vector-stores-files

PR #468, opened 2024-05-02
feat(api): vector store endpoints

### LINKED ISSUE #421: feat(api): Vector Stores endpoints
### User Story

**As a** LeapfrogAI Developer 
**I want** an OpenAI compliant Vector Stores API
**So that** I can interact with the vector store via the API

### Acceptance Criteria

Implement and test the following endpoints:
- [ ] Create vector store
- [ ] List vector store
- [ ] Retrieve vector store
- [ ] Modify vector store
- [ ] Delete vector store

### Additional context
- https://platform.openai.com/docs/api-reference/vector-stores

gphorvath:
Contains:

- Supabase Migration for Vector Store, Vector Content (where the vectors are), Vector Store Files (join between Vector Store and Files). Also adds functions for calculating size of vector content and doing similarity search
- CRUD operations for Vector Stores and Vector Store Files
- Index/Query operations for RAG
- A leapfrogai endpoint for testing RAG.
- Integration test for Vector Stores
- Fixes api logging using standard python logger
- Automatic vector store expiry when specified

Requires:

- Supabase w/ auth (or db ops fail)
- Embeddings backend (or indexing breaks)

closes #421 #422 #490


supabase/migrations/20240502193159_init-vector-store.sql
```diff
@@ -0,0 +1,68 @@
...
+    vector_store_id uuid references vector_store_objects (id) on delete cascade,
+    file_id uuid references vector_store_file_objects (id) on delete cascade,
+    content text, -- corresponds to Document.pageContent
+    metadata jsonb, -- corresponds to Document.metadata
+    embedding vector (768) -- Instructor-XL produces 768-dimensional embeddings
+  );
+
+-- Create a function to search for documents
+create function match_vectors (
+  query_embedding vector (768), -- Instructor-XL produces 4096-dimensional embeddings
```
gphorvath, line 45: 768 length

gphorvath:
Starting to work on storing and querying across vectors. Noticed that vecs and Langchain's Supabase client don't support async calls. Might need to use something like psycopg.

src/leapfrogai_api/routers/openai/files.py
```diff
@@ -7,17 +7,21 @@
 from leapfrogai_api.data.crud_file_object import CRUDFileObject
 from leapfrogai_api.data.crud_file_bucket import CRUDFileBucket
 from leapfrogai_api.routers.supabase_session import Session
+from leapfrogai_api.backend.rag.document_loader import supported_mime_type
 
 router = APIRouter(prefix="/openai/v1/files", tags=["openai/files"])
 
 
 @router.post("")
 async def upload_file(
```
gphorvath, line 16: Do we want file size limits on upload?

src/leapfrogai_api/backend/rag/document_loader.py
```diff
@@ -0,0 +1,95 @@
...
+from langchain_community.document_loaders import (
+    CSVLoader,
+    Docx2txtLoader,
+    PyPDFLoader,
+    TextLoader,
+    UnstructuredHTMLLoader,
+    UnstructuredMarkdownLoader,
+)
+from langchain_core.documents import Document
+from langchain_text_splitters import RecursiveCharacterTextSplitter
```
gphorvath, line 14: Follow-on work would be to use semantic splitting with our LLM backend.

src/leapfrogai_api/README.md
```diff
@@ -5,6 +5,7 @@ A mostly OpenAI compliant API surface.
 ## Requirements
 
 - Supabase
+- Libreoffice ([Unstructured dependency via LangChain](https://python.langchain.com/docs/integrations/providers/unstructured/) for docx parsing)
```
alekst23, line 8: Do we actually implement this somewhere, or is it just a dependency library?

src/leapfrogai_api/backend/leapfrogai_embeddings.py
```diff
@@ -0,0 +1,76 @@
...
+
+        Returns:
+            model: The embeddings model.
+
+        Raises:
+            ValueError: If the embeddings model is not found.
+        """
+        model = get_model_config().get_model_backend(model=model_name)
+        if model is None:
+            raise ValueError("Embeddings model not found.")
```
alekst23, line 74:
```suggestion
        if not model := get_model_config().get_model_backend(model=model_name)
            raise ValueError("Embeddings model not found.")
```

src/leapfrogai_api/README.md
gphorvath: This is a dependency of Unstructured a library included in Langchain for parsing various types to document.

src/leapfrogai_api/data/crud_base.py
```diff
@@ -21,7 +21,8 @@ def __init__(
     async def create(self, object_: ModelType) -> ModelType | None:
         """Create new row."""
         dict_ = object_.model_dump()
-        del dict_["id"]  # Ensure this is created by the database
+        if dict_["id"] == "":  # for some endpoints an id may be provided
```
alekst23, line 24: If this property is not always present, and we want to remove it regardless of it's value, we should use this pattern:
```python
if "id" in dict_:
    del dict_["id"]
```

alekst23 (changes requested):

supabase/migrations/20240502193159_init-vector-stores.sql
```diff
@@ -0,0 +1,104 @@
...
+#variable_conflict use_column
+begin
+  return query
+  select
+    id,
+    vector_store_id,
+    file_id,
+    content,
+    metadata,
+    1 - (vector_store.embedding <=> query_embedding) as similarity
```
alekst23, line 96: Depending on how the query management is handled for the `<=>` operator, it is quite possible that this SQL will perform the cosine similarity search twice, once for the column data, and once for the order by operation. 

There's a couple ways we can solve this. One is rewriting the query with a common table expression. The other is making sure that the query is index and only runs once per call. I think we may need to create a new index for this operator regardless, actually. Still doing some research into this now. 

Given that a lot of example have this type of syntax without CTE, I imagine the indexing step resolves this issue.

alekst23: I am 80% sure that postgres will optimize this query. However, I want to follow up with some testing later.

src/leapfrogai_api/backend/rag/document_loader.py
```diff
@@ -0,0 +1,94 @@
...
+    return await text_splitter.atransform_documents(docs)
+
+
+async def embed_chunks(chunks: list[Document]) -> list[list[float]]:
+    """Embed chunks into a document."""
+
+    model = get_model_config().get_model_backend(os.getenv("DEFAULT_EMBEDDINGS_MODEL"))
+
+    if model is None:
+        raise ValueError("Embeddings model not found.")
```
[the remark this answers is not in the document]
gphorvath: fixed.

src/leapfrogai_api/data/crud_base.py
gphorvath: I think that makes sense. In the case of vector_store_file I discovered the need to provide the ID, so I changed this.

src/leapfrogai_api/data/supabase_vector_store.py
```diff
@@ -0,0 +1,185 @@
...
+
+    """
+
+    async def __ainit__(  # pylint: disable=arguments-differ
+        self,
+        jwt: str,
+        embedding: Embeddings | None = None,
+        table_name: str = "vector_store",
+        chunk_size: int = 500,
+        query_name: str = "match_vectors",
```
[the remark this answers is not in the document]
CollectiveUnicorn: I've hardcoded these, so they should be good now

src/leapfrogai_api/backend/rag/leapfrogai_embeddings.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        model = await self._get_model()
+        request = lfai.EmbeddingRequest(inputs=texts)
+        response = await create_embeddings(model=model, request=request)
+
+        list_of_embeddings = [data.embedding for data in response.data]
+
+        return list_of_embeddings
+
+    async def aembed_query(self, text: str) -> list[float]:
```
alekst23, line 33: We should try to avoid duplicate code like in functions `aembed_query` and `aembed_document`, which perform similar functionality. Maybe `aembed_documents` should call `aembed_query` iteratively?

CollectiveUnicorn:
Set to debug log level by default

src/leapfrogai_api/backend/rag/document_loader.py
```diff
@@ -0,0 +1,96 @@
...
+    "application/pdf": PyPDFLoader,
+    "text/plain": TextLoader,
+    "text/html": UnstructuredHTMLLoader,
+    "text/csv": CSVLoader,
+    "text/markdown": UnstructuredMarkdownLoader,
+    "application/msword": Docx2txtLoader,
+}
+
+
+async def supported_mime_type(mime_type: str) -> bool:
```
CollectiveUnicorn, line 32: Make not async

CollectiveUnicorn (changes requested):
Set to debug log-level by default wherever uvicorn is used

src/leapfrogai_api/routers/openai/vector_stores.py
```diff
@@ -0,0 +1,299 @@
...
+    session: Session,
+) -> ListVectorStoresResponse:
+    """List all the vector stores."""
+
+    crud_vector_store = CRUDVectorStore(db=session)
+    crud_response = await crud_vector_store.list()
+
+    return ListVectorStoresResponse(
+        object="list",
+        data=crud_response or [],
```
YrrepNoj, line 96: Does your IDE give warnings on the type hinting here for you?

The `ListVectorStoresResponse.data` object is defined as `list[VectorStore]` but the list we're passing in here is a `list[AuthVectorStore]`

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -0,0 +1,98 @@
...
+                )
+
+                ids = await vector_store_client.aadd_documents(
+                    documents=chunks,
+                    vector_store_id=vector_store_id,
+                    file_id=file_id,
+                )
+
+                if len(ids) == 0:
+                    vector_store_file.status = "failed"
```
YrrepNoj, line 96: Can we make the status string values an ENUM or statically defined somewhere in this file/class instead of using magic-strings?

This way, the places validating the status can compare against the static types instead of validating against more magic-strings.

src/leapfrogai_api/routers/openai/vector_stores.py
```diff
@@ -0,0 +1,299 @@
...
+                    new_vector_store.file_counts.completed += 1
+                elif response.status == "failed":
+                    new_vector_store.file_counts.failed += 1
+                elif response.status == "in_progress":
+                    new_vector_store.file_counts.in_progress += 1
+                elif response.status == "cancelled":
+                    new_vector_store.file_counts.cancelled += 1
+                new_vector_store.file_counts.total += 1
+
+        new_vector_store.status = "completed"
```
YrrepNoj, line 179: Nit; We should probably be checking against static variables for the status instead of comparing/setting them to string literals.

src/leapfrogai_api/routers/openai/vector_stores.py
```diff
@@ -0,0 +1,299 @@
...
+) -> VectorStoreDeleted:
+    """Delete a vector store."""
+
+    crud_vector_store = CRUDVectorStore(db=session)
+
+    vector_store_deleted = await crud_vector_store.delete(id_=vector_store_id)
+    return VectorStoreDeleted(
+        id=vector_store_id,
+        object="vector_store.deleted",
+        deleted=bool(vector_store_deleted),
```
YrrepNoj, line 213: Question; `vector_store_deleted` was already a `bool` type. Are we trying to do something here that I might be missing?

YrrepNoj (changes requested):

src/leapfrogai_api/data/crud_vector_store_file.py
```diff
@@ -0,0 +1,99 @@
...
+            .execute()
+        )
+
+        _, response = data
+
+        if response:
+            return self.model(**response[0])
+        return None
+
+    async def delete(self, vector_store_id: str, file_id: str) -> bool:  # pylint: disable=arguments-differ
```
YrrepNoj, line 86: The pylint annotations here feels like a smell. We are overloading an inherited method with different arguments. Is there a different we can architect this?

At the very least, we should leave a descriptive comment justifying why we are overloading these methods in this way.

src/leapfrogai_api/data/crud_base.py
```diff
@@ -18,7 +18,8 @@ def __init__(self, db: AsyncClient, model: type[ModelType], table_name: str):
     async def create(self, object_: ModelType) -> ModelType | None:
         """Create new row."""
         dict_ = object_.model_dump()
-        del dict_["id"]  # Ensure this is created by the database
+        if "id" in dict_ and not dict_.get("id"):
+            del dict_["id"]
         del dict_["created_at"]  # Ensure this is created by the database
```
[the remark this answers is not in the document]
gphorvath: done

src/leapfrogai_api/routers/openai/files.py
```diff
@@ -19,6 +20,12 @@ async def upload_file(
 ) -> FileObject:
     """Upload a file."""
 
+    if not supported_mime_type(request.file.content_type):
+        raise HTTPException(
+            status_code=405,
```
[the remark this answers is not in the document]
gphorvath: agreed, fixed.

src/leapfrogai_api/backend/rag/document_loader.py
```diff
@@ -0,0 +1,96 @@
...
+    "application/pdf": PyPDFLoader,
+    "text/plain": TextLoader,
+    "text/html": UnstructuredHTMLLoader,
+    "text/csv": CSVLoader,
+    "text/markdown": UnstructuredMarkdownLoader,
+    "application/msword": Docx2txtLoader,
+}
+
+
+def supported_mime_type(mime_type: str) -> bool:
```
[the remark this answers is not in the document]
gphorvath: changed

src/leapfrogai_api/routers/openai/vector_stores.py
gphorvath: This was holdover from when these endpoints return something other than a bool. We simplified that, so removing this bool check.

src/leapfrogai_api/routers/openai/vector_stores.py
gphorvath: changed to enum

src/leapfrogai_api/backend/rag/index.py
gphorvath: changed to enum

gphorvath: now this enummed

packages/api/zarf.yaml
```diff
@@ -20,6 +20,8 @@ variables:
     description: "Flag to expose the OpenAPI schema for debugging."
   - name: HOSTED_DOMAIN
     default: "uds.dev"
+  - name: DEFAULT_EMBEDDINGS_MODEL
```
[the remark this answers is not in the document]
gphorvath: This prevents the need for the caller to provide a field not specified in the OpenAI request model. When files are added to the Vector Store, the request type doesn't require an embeddings model be provided, but it will need to calculate the embeddings. For reference: https://platform.openai.com/docs/api-reference/vector-stores-files/createFile

I'll let you resolve this comment if you think this acceptable.

src/leapfrogai_api/backend/rag/document_loader.py
```diff
@@ -0,0 +1,96 @@
...
+        separators=separators,
+    )
+
+    return await text_splitter.atransform_documents(docs)
+
+
+async def embed_chunks(chunks: list[Document]) -> list[list[float]]:
+    """Embed chunks into a document."""
+
+    model = get_model_config().get_model_backend(os.getenv("DEFAULT_EMBEDDINGS_MODEL"))
```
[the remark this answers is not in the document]
gphorvath: This function was moved to indexing, removed the redundant function here.

src/leapfrogai_api/routers/openai/vector_stores.py
gphorvath: In theory this would pare down anything from the superclass that wasn't needed. That said, I change crud_vector_store to return regular VectorStore object since we aren't using the additional user data anywhere.

src/leapfrogai_api/data/crud_vector_store_file.py
gphorvath: I added a little more commentary here. The base class doesn't allow for two ids (vector_store and file) to be passed, so we needed to make an exception here.

src/leapfrogai_api/routers/openai/embeddings.py
```diff
@@ -33,6 +33,8 @@ async def embeddings(
 
     if isinstance(req.input, str):
         request = lfai.EmbeddingRequest(inputs=[req.input])
+    elif isinstance(req.input, list):
+        request = lfai.EmbeddingRequest(inputs=req.input)
```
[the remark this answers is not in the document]
gphorvath: fixed.

src/leapfrogai_api/data/crud_base.py
```diff
@@ -18,7 +18,8 @@ def __init__(self, db: AsyncClient, model: type[ModelType], table_name: str):
     async def create(self, object_: ModelType) -> ModelType | None:
         """Create new row."""
         dict_ = object_.model_dump()
-        del dict_["id"]  # Ensure this is created by the database
+        if "id" in dict_ and not dict_.get("id"):
+            del dict_["id"]
```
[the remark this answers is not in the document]
gphorvath: In some cases the ID is provided, so we only delete the empty ID

tests/integration/api/test_files.py
```diff
@@ -125,3 +124,19 @@ def test_get_nonexistent():
...
+
+
+def test_invalid_file_type():
+    """Test creating uploading an invalid file type."""
+
+    file_path = "../../../tests/data/0min12sec.wav"
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    relative_file_path = os.path.join(dir_path, file_path)
+
+    with pytest.raises(HTTPException):
```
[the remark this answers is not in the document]
gphorvath: fixed

src/leapfrogai_api/data/crud_vector_store_file.py
YrrepNoj: We control the base class we're using here though right? We could add an optional parameter of `eqMap: typing.Dict[str, str]` where the dictionary key is the name of the column and the value is the value we are setting the "equal to" clause to.

packages/api/zarf.yaml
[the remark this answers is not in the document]
YrrepNoj: Hmm, I see we're trying to conform to the OpenAI request model, but we might need to do something different here. Our API should be as stateless as possible, it shouldn't have to know what other things exist or will exist later. The only caveat to this is that our API is watching for configmaps to see what valid model names are, but that is done in a way that is fully dynamic and not 'built into' the deployment.


The way this is currently implemented, you can only set the embeddings_model during the deploy time of the API. Users (or apps) would not be able to specify a different model, they would instead have to wait for someone to change the environment variable in their deployment manifest and redeploy the deployment.

src/leapfrogai_api/routers/openai/vector_stores.py
```diff
@@ -0,0 +1,297 @@
...
+@router.get("/{vector_store_id}/files/{file_id}")
+async def retrieve_vector_store_file(
+    vector_store_id: str,
+    # file_id: str,
+    session: Session,
+):  # -> VectorStoreFile:
+    """Retrieve a file in a vector store."""
+
+    crud_vector_store_file = CRUDVectorStoreFile(db=session)
+    return await crud_vector_store_file.get(filters={"vector_store_i": vector_store_id})
```
CollectiveUnicorn, line 269:
```suggestion
    return await crud_vector_store_file.get(filters={"vector_store_id": vector_store_id})
```

CollectiveUnicorn (changes requested):

src/leapfrogai_api/backend/rag/document_loader.py
```diff
@@ -0,0 +1,71 @@
...
+        "\uff0c",  # Full width comma
+        "\u3001",  # Ideographic comma
+        "\uff0e",  # Full width full stop
+        "\u3002",  # Ideographic full stop
+        "",
+    ]
+
+    text_splitter = RecursiveCharacterTextSplitter(
+        # Set a really small chunk size, just to show.
+        chunk_size=500,
```
YrrepNoj, line 64: Is there anything specifically relevant to the `500` value? 'Just to show' what?

src/leapfrogai_api/routers/openai/vector_stores.py
```diff
@@ -0,0 +1,305 @@
...
+        new_vector_store = VectorStore(
+            id=vector_store_id,
+            bytes=old_vector_store.bytes,  # Automatically calculated by DB
+            created_at=old_vector_store.created_at,
+            file_counts=old_vector_store.file_counts,
+            last_active_at=old_vector_store.last_active_at,  # Update after indexing files
+            metadata=getattr(request, "metadata", old_vector_store.metadata),
+            name=getattr(request, "name", old_vector_store.name),
+            object="vector_store",
+            status="in_progress",
```
YrrepNoj, line 143: Could this use the new `VectorStoreStatus` enum?

gphorvath: Good catch. Fixed.

src/leapfrogai_api/backend/rag/document_loader.py
gphorvath: I updated the comment. Really what I mean is that these parameters are just some defaults suggested by Langchain and we may need to tune these once we start getting an idea how they perform

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -0,0 +1,113 @@
...
+        self.embeddings = embeddings_type()
+
+    async def index_file(self, vector_store_id: str, file_id: str) -> VectorStoreFile:
+        """Index a file into a vector store."""
+        crud_vector_store_file = CRUDVectorStoreFile(db=self.db)
+
+        if await crud_vector_store_file.get(
+            filters={"vector_store_id": vector_store_id, "id": file_id}
+        ):
+            raise FileAlreadyIndexedError("File already indexed")
```
alekst23, line 42: How necessary is this custom exception class? It's a little redundant and counter-pattern to what we have elsewhere.

Would it be better to do something like this:
```python
raise HTTPException(
            status_code=status.HTTP_400_BAD_REQUEST, detail="File already indexed"
        ) 
```

gphorvath: That is totally fair. I changed it!

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -0,0 +1,113 @@
...
+                    id=file_id,
+                    created_at=0,
+                    last_error=LastError(
+                        message="No text found in file", code="parsing_error"
+                    ),
+                    object="vector_store.file",
+                    status=VectorStoreFileStatus.FAILED.value,
+                    vector_store_id=vector_store_id,
+                )
+                return await crud_vector_store_file.create(object_=vector_store_file)
```
alekst23, line 70: Is there a reason we want to store an empty file that failed to process?

It feels like we can raise an error here.

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -0,0 +1,113 @@
...
+                )
+
+                ids = await vector_store_client.aadd_documents(
+                    documents=chunks,
+                    vector_store_id=vector_store_id,
+                    file_id=file_id,
+                )
+
+                if len(ids) == 0:
+                    vector_store_file.status = VectorStoreFileStatus.FAILED.value
```
alekst23, line 97: Should we raise an exception here as well?

src/leapfrogai_api/pyproject.toml
```diff
@@ -11,10 +11,16 @@ dependencies = [
     "openai == 1.21.1",
     "uvicorn >= 0.23.2",
     "pydantic >= 2.0.0",
+    "docx2txt >= 0.8",
     "python-multipart >= 0.0.7", #indirect dep of FastAPI to receive form data for file uploads
     "watchfiles >= 0.21.0",
     "leapfrogai_sdk",
-    "supabase-py-async",
+    "supabase-py-async >= 2.5.6",
+    "langchain >= 0.2.1",
```
alekst23, line 19: Is there something we use langchain for besides `from langchain_core.embeddings.fake import FakeEmbeddings`?

We may want to rethink that approach if that's the case. Langchain has, in my past experience, caused huge dependency chain issues. It will attempt to download all sorts of libs and maybe models unpredictably.

gphorvath: The short answer is yes.

We use it in the document loader and indexing service right now. We aren't currently using any features that should pull in models, but we are aware of some of them (like the context splitter).

packages/api/zarf.yaml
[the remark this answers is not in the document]
gphorvath: @CollectiveUnicorn put in some issues to cover this, resolving for now.

tests/pytest/leapfrogai_api/test_api.py
```diff
@@ -134,6 +134,40 @@ def test_routes():
         ("/openai/v1/assistants/{assistant_id}", "retrieve_assistant", ["GET"]),
         ("/openai/v1/assistants/{assistant_id}", "modify_assistant", ["POST"]),
         ("/openai/v1/assistants/{assistant_id}", "delete_assistant", ["DELETE"]),
+        ("/openai/v1/vector_stores", "create_vector_store", ["POST"]),
+        ("/openai/v1/vector_stores", "list_vector_stores", ["GET"]),
+        ("/openai/v1/vector_stores", "list_vector_stores", ["GET"]),
```
alekst23, line 139: Duplicate?

src/leapfrogai_api/routers/openai/vector_stores.py
```diff
@@ -0,0 +1,305 @@
...
+    vector_store_id: str,
+    session: Session,
+) -> VectorStore | None:
+    """Retrieve a vector store."""
+
+    crud_vector_store = CRUDVectorStore(db=session)
+    return await crud_vector_store.get(filters=FilterVectorStore(id=vector_store_id))
+
+
+@router.post("/{vector_store_id}")
```
alekst23, line 113: Should the update/modify methods be "PATCH"? I think that's the more RESTFUL practice.

tests/pytest/leapfrogai_api/test_api.py
gphorvath: yes!

src/leapfrogai_api/routers/openai/vector_stores.py
gphorvath: That makes sense but the OpenAI spec specifies POST. Do you think we'll end up with a mismatch when using OpenAI libraries if we switch to PATCH?

alekst23: Yeah if we end up trying to do any sort of procedural comparison between the two, as you had suggested, we would end up with issues.

alekst23 (approved):

CollectiveUnicorn (approved):

YrrepNoj (approved):

PR #487, opened 2024-05-08
update llama-cpp-python version

jalling97:
- Updates the version of llama-cpp-python to `0.2.70` to prevent cuda errors from appearing in CPU only deployments
- Adds an extra link to the wheel command for the `Makefile` as a way of specifying a pre-built CPU wheel


justinthelaw (dismissed):

gphorvath (dismissed):

Makefile
```diff
@@ -67,7 +67,9 @@ build-ui: ## Build the leapfrogai_ui container and Zarf package
 
 setup-llama-cpp-python-deps: sdk-wheel ## Download the wheels for the optional 'llama-cpp-python' dependencies
 	-rm packages/llama-cpp-python/build/*.whl
-	python -m pip wheel packages/llama-cpp-python -w packages/llama-cpp-python/build --find-links=${SDK_DEST}
+
+	## The external link is needed to pull a pre-compiled cpu wheel for llama-cpp-python
+	python -m pip wheel packages/llama-cpp-python -w packages/llama-cpp-python/build --find-links=${SDK_DEST} --extra-index-url https://abetlen.github.io/llama-cpp-python/whl/cpu
```
YrrepNoj, line 72: Is the `--extra-index-url` necessary? I tried running this target with and without this flag and the .whl binaries didn't show any diff with the `diff` command.

YrrepNoj:
@jalling97 could you write up an issue (and link it to this PR) that describes the cuda errors you were running into and how someone could try to replicate them? Thanks!

Makefile
jalling97: The newest releases of llama-cpp-python direct that we can download a pre-built cpu wheel from this url instead. In my mind this ensures that we'll always end up with a cpu-only build, but I get that this further muddles an already complex build process

justinthelaw: If we ever decide to have GPU + CPU offloading support, the pre-built llama-cpp-python .whl will be super useful. It'd be good to keep that argument there as documentation for a future refactor. It also adds more clarifty to the fact that llama-cpp-python is only CPU right now in our repository.

The reason is our old GPU llama-cpp-python Dockerfile used the NVIDIA CUDA container to pre-build the llama-cpp-python .whl before moving it into our python container, which was a really odd practice and we should not be maintaining for ourselves anyway. Having it done by the upstream developer is more consistent and we have more access to different architectures, like Metal ARM64, ARM64, AMD64 etc.

justinthelaw (approved):

YrrepNoj (approved):

Issue #463, opened 2024-04-30
feat(api): return finish_reason field for chat completions and chat completions streaming


andrewrisse:
### User Story

**As a** frontend developer
**I want** to reference the finish_reason field in a chat completion
**So that** I can implement a continue generation feature

Issue #388 requires this field in order to be implemented.

The API is currently returning null for the finish_reason field. 
Update to return IAW: https://platform.openai.com/docs/guides/text-generation/chat-completions-response-format

### Describe alternatives you've considered
We can also look for a termination character in the string, but this is more hacky and the finish_reason field exists for this purpose.

Issue #431, opened 2024-04-23
Determine why VLLM will not startup with decorators, replace non-decorator deployment


CollectiveUnicorn:
### Describe what should be investigated or refactored
When updating to confz for VLLM, an issue was identified where VLLM would not startup if the leapfrogai-api decorators were used (like @LLM). 

Once a root cause is identified, replace the current implementation with the decorators.

### Links to any relevant code
https://github.com/defenseunicorns/leapfrogai/pull/331

### Additional context
This only occurs when doing a k8s deployment, the issue does not occur when using docker or running locally.

PR #493, opened 2024-05-09
Return finish reason field & VLLM Refactor

CollectiveUnicorn:
When performing a completion or chat completion return the `finish_reason` according to the api spec.
* Adds finish reason in completion/chat completion that mirrors the openai spec
* Adds token usage in completion/chat completion responses that mirrors the openai spec

Closes https://github.com/defenseunicorns/leapfrogai/issues/463
Closes https://github.com/defenseunicorns/leapfrogai/issues/431


CollectiveUnicorn:
Conforming to the specs for token length will require updating all of the backends. It assumes that there's the token counting function. TBD if doing that in this commit.

justinthelaw (commented):
Just a nit, but are there any additional tests to check the token count field using the repeater model? If this should wait for other components to implement the consumption and display of the resulting token count, then recommend adding another issue to track that, if not done already.

justinthelaw (approved):
Minor nitpick for future LFAI members: we should add release-please or use renovate to change version numbers in our CI workflows and artifacts (Makefile, or in the future UDS tasks and task runner)

justinthelaw (dismissed):

src/leapfrogai_sdk/llm.py
```diff
@@ -40,6 +42,58 @@ def LLM(_cls):
     if not hasattr(_cls, "generate"):
         raise ValueError("LLM class requires a generate method")
 
+    if not hasattr(_cls, "count_tokens"):
+        raise ValueError("LLM class requires a count_tokens method")
+
+    def create_chat_completion_response(
+        text: str,
+        finish_reason: str = None,
```
YrrepNoj, line 50: I'm a little surprised our linter isn't catching this, but a `str` is not an object and therefor cannot be assigned `None`.

I think this might make more sense as an empty string for the zero-value?

src/leapfrogai_sdk/llm.py
```diff
@@ -107,17 +198,54 @@ async def Complete(
...
+
+            prompt_token_count: int = await self.count_tokens(request.prompt)
+
+            return create_completion_response(
+                content, finish_reason, prompt_token_count, completion_token_count
+            )
 
         async def CompleteStream(
             self, request: CompletionRequest, context: GrpcContext
         ) -> Generator[CompletionResponse, Any, Any]:
```
YrrepNoj, line 216: I'm noticing that this is an async (streamed) response but we're using `Generator` instead of `AsyncGenerator`.


Does `AsyncGenerator` make more sense here? Does now seem like a good time to update that?

src/leapfrogai_sdk/llm.py
```diff
@@ -107,17 +198,54 @@ async def Complete(
...
+
+                last_delta = text_chunk
+
+            if last_delta:
+                response_str += last_delta
+
+            completion_token_count: int = await self.count_tokens(response_str)
+
+            if completion_token_count < request.max_new_tokens:
+                finish_reason = "stop"
```
YrrepNoj, line 238: What are your thoughts on turning these finish_reasons into an Enum class instead of evaluating them as magic strings?

src/leapfrogai_sdk/llm.py
```diff
@@ -40,6 +42,58 @@ def LLM(_cls):
...
+                completion_tokens=completion_tokens,
+                total_tokens=prompt_tokens + completion_tokens,
+            )
+
+        response: ChatCompletionResponse = ChatCompletionResponse(
+            choices=[choice], usage=usage
+        )
+
+        if finish_reason:
+            response.choices[0].finish_reason = finish_reason
```
YrrepNoj, line 70: Question; Why put this assignment behind the conditional? 

From my understanding, `responses.choices[0].finish_reason` will always be the 0-value at this point right?

YrrepNoj (changes requested):

tests/pytest/leapfrogai_api/test_api.py
```diff
@@ -197,6 +210,17 @@ def test_stream_chat_completion():
...
+                    # parse finish reason
+                    assert "finish_reason" in choices[0]
+                    assert "stop" == choices[0].get("finish_reason")
+                    # parse usage data
+                    response_usage = stream_response.get("usage")
+                    prompt_tokens = response_usage.get("prompt_tokens")
+                    completion_tokens = response_usage.get("completion_tokens")
+                    total_tokens = response_usage.get("total_tokens")
+                    assert prompt_tokens == 1
+                    assert completion_tokens == 2
```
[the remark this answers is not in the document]
YrrepNoj: We discussed this in person; The repeater model just returns hardcoded values for the purpose of testing. 👍

src/leapfrogai_sdk/llm.py
CollectiveUnicorn: It just converts it into an empty string, I'll replace it with an explicit empty string.

packages/repeater/chart/Chart.yaml
```diff
@@ -16,7 +16,7 @@ type: application
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 # x-release-please-start-version
-version: 0.7.2
+version: 0.7.3
```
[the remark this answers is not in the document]
CollectiveUnicorn: Alright, I will go through the standard release channels prior to updating test dependencies. I can decrement these and do that separately.

src/leapfrogai_sdk/llm.py
CollectiveUnicorn: That's fine, I replaced them with an enum containing the following values:
- NONE
- STOP
- LENGTH

src/leapfrogai_sdk/llm.py
CollectiveUnicorn: You're right, this is unnecessary and have removed the conditional.

src/leapfrogai_sdk/llm.py
CollectiveUnicorn: I'll test it and see if it works. If it does then I'll leave the references as AsyncGenerators

CollectiveUnicorn: The async change appears to work

packages/repeater/chart/Chart.yaml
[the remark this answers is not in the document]
CollectiveUnicorn: All the version have been reverted back to 0.7.2 and the pipeline workflow now builds a dev version of repeater for the test

src/leapfrogai_sdk/llm.py
CollectiveUnicorn: I take that back, there appears to be some issue

CollectiveUnicorn: The issues has been resolved, it was just a mismatch in the number of parameters the AsyncGenerator was expecting vs the Generator. It works fine now.

justinthelaw (approved):

YrrepNoj (approved):

gphorvath (approved):

PR #496, opened 2024-05-10
chore: Supabase migrations

### LINKED ISSUE #481: Update API and UI Zarf package to run database migrations
### User Story

(Assuming we have a Postgres instance existing in the cluster) When we run `zarf package deploy zarf-package-api-.....tar.zst` we should expect the database migrations to have been applied.

Two possible approaches for this is to perform the migration in-cluster with a k8s job or on the host machine as a `Zarf.action.onDeploy` action.

YrrepNoj:
This PR takes the various `*.sql` files that we have and sets the schema of the tables during the supabase-postgres chart deployment

Closes #481


packages/supabase/bitnami-values.yaml
```diff
@@ -178,10 +178,23 @@ postgresql:
...
   primary:
     resourcesPreset: "none"
     podLabels:
       sidecar.istio.io/inject: "false"
+    initdb:
+      scripts:
+        0000000000000000_migrate.sh: |
+          ###ZARF_VAR_MIGRATION_SCRIPT###
+
+        20240322174520_api_sql_schema.sql: |
```
justinthelaw, line 191: Is the intent that we will be manually entering migrations here, in temporal order, for the container to run? Also see my later comment.

packages/supabase/zarf.yaml
```diff
@@ -21,6 +21,19 @@ constants:
     value: "https://keycloak.admin.uds.dev/realms/uds"
 
 variables:
+  - name: UI_SQL_SCHEMA
```
justinthelaw, line 24: Should these not be create-time, constants? The delivery engineer _could_ potentially point to a modified SQL script, but in reality these migrations should be controlled by the package and bundle creators.

Additionally, since the place in the manifests where these are replaced is static, yet they are named with a temporal ID, it seems like we would want to instead import an extra-values file where we list the migrations and track them in that separate file.

justinthelaw (commented):
Some clarifying questions/ideas!

packages/supabase/zarf.yaml
YrrepNoj: The way this works now, they are build set and managed at build-time, like you said they should be :)

packages/supabase/bitnami-values.yaml
YrrepNoj: Yess-ish.

The migrations defined here will get mounted into the supabase container and run in sorted order based on the filenames.

justinthelaw (approved):

packages/supabase/migrations/api_sql_schema.sql
gphorvath, line 1: I'm still wondering if it makes sense to date these migrations so that we can easily infer order?

packages/supabase/migrations/ui_sql_schema.sql
gphorvath, line 1: I'm still wondering if it makes sense to date these migrations so that we can easily infer order?

packages/supabase/migrations/api_sql_schema.sql
YrrepNoj: Added a date to the filename.

packages/supabase/migrations/ui_sql_schema.sql
YrrepNoj: Added a timestamp to the filename prefix.

justinthelaw (dismissed):

gphorvath (approved):

PR #505, opened 2024-05-14
Add supabase to release workflow

YrrepNoj:
This PR adds supabase to the release workflow so that an amd64 and arm64 zarf package will be published whenever a tag is pushed to this repo.

This PR also adds `milestoned` as an event trigger to our workflows. This allows us to trigger workflows on PRs that don't normally trigger a workflow (such as when the author is  `github-actions`.


barronstone (approved):
By my neanderthal understanding of release workflows, this looks legit to me 👍

PR #507, opened 2024-05-14
chore: mac deployments

alekst23:
This PR adds support for building and deploy packages on Macs.

The issue arrises because the current build process is configured to build the packages on the local machine, and then copy the package for installation to a Docker image. Doing this on Mac causes the packages to be built for the local Mac architecture, and these file versions do not work in the Docker Linux container.

We address the issue in this PR by moving the build process inside Docker:
- Makefile now copies all required files to the package folder first
- Dockerfile for each package builds the package dependencies and installs the package itself inside the `builder` layer
- We add an optional ARCH parameter for the `make` command, which defaults to `amd64`, and can be set to `arm64` for building on Mac
- We add an optional REG_PORT parameter to `make`, to allow the `registry` container to change port if `5000` is already taken on the local system

Additional Notes:
- We change `lfaiui` to `leapfrogai-ui` in the `ui` package


alekst23:
I need to test this locally outside docker, but at this point the build and uds deploy should be working on mac.

uds-bundles/dev/cpu/uds-bundle.yaml
```diff
@@ -53,3 +53,8 @@ packages:
   - name: leapfrogai-ui
     path: ../../../packages/ui/
     ref: dev
+
+  # Supabase backend for the UI and API to interface with Postgresql
+  - name: supabase
```
YrrepNoj, line 58: I think this might be a copy/paste artifact? There is a `-name: supabase` key ~20 lines above.

YrrepNoj (changes requested):
One small change request, but other than that this PR is looking good to me!

YrrepNoj:
@alekst23 Can you update the `vllm` Maketarget and Dockerfile so that it will conform with the rest of our builds?

I know vllm isn't MacOS compatible so it wasn't in the original scope of what you were doing here, but I think it's important enough to keep our builds consistent.

Makefile
```diff
@@ -31,123 +32,118 @@ gen-python: ## Generate the protobufs for the OpenAI typing within the leapfroga
...
 	## Build the Zarf package
-	uds zarf package create packages/ui -o packages/ui --registry-override=ghcr.io=localhost:5000 --insecure --set IMAGE_VERSION=${LOCAL_VERSION} --confirm
-
+	uds zarf package create packages/ui -o packages/ui --registry-override=ghcr.io=localhost:${REG_PORT} --insecure --set IMAGE_VERSION=${LOCAL_VERSION} --confirm
 
-setup-llama-cpp-python-deps: sdk-wheel ## Download the wheels for the optional 'llama-cpp-python' dependencies
-	-rm packages/llama-cpp-python/build/*.whl
-
-	## The external link is needed to pull a pre-compiled cpu wheel for llama-cpp-python
-	python -m pip wheel packages/llama-cpp-python -w packages/llama-cpp-python/build --find-links=${SDK_DEST}
```
YrrepNoj, line 76: The llama-cpp-python Dockerfile was taking a 'shortcut' to see if a model existed in the `build/` directory to try to bypass the model-download script. If the `build/` directory doesn't exist than our Dockerfile will run into an error when trying to build.

The Dockerfile is trying to do the [copy here](https://github.com/defenseunicorns/leapfrogai/blob/c9b5adb0e84134476c9776dcd8a43d72cb32b49f/packages/llama-cpp-python/Dockerfile#L16) and the [workflow is currently failing](https://github.com/defenseunicorns/leapfrogai/actions/runs/9134869017/job/25197381687?pr=507#step:12:156) because the `build/` directory doesn't exist.

I think we can get away with removing this shortcut, but I wouldn't be opposed to just adding another conditional check here either.

uds-bundles/dev/cpu/uds-bundle.yaml
alekst23: Indeed! It was from a rebase that I did not clean up well.

justinthelaw (approved):
Everything, with the exception of @YrrepNoj comments, looks good to me! 

Future reference: the copying over of `src`  into the build context's directory is an interesting move to get rid of the relative import within the `pyproject.toml`. I think a clean-up of these, not just a `.gitignore`, would be a nice touch to the `Makefile`. (understanding that this isn't your original pattern, but an already present one)

Thanks for all the testing and improvements!

gphorvath:
Closing in favor of https://github.com/defenseunicorns/leapfrogai/pull/659

PR #529, opened 2024-05-16
Add RAG Evals ADR

jalling97:
Adds an ADR for what framework we will use to handle RAG evaluations.


adr/0004-rag-eval-framework.md
```diff
@@ -0,0 +1,59 @@
+# LeapfrogAI RAG Evaluations
+
+## Table of Contents
+
+- [LeapfrogAI RAG Evaluations](#leapfrogai-rag-evaluations)
+  - [Table of Contents](#table-of-contents)
+  - [Status](#status)
+  - [Background](#background)
```
CollectiveUnicorn, line 8: This doesn't go anywhere

CollectiveUnicorn (changes requested):

CollectiveUnicorn (approved):

adr/0004-rag-eval-framework.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Alternatives
+**RAGAS**: A more lightweight alternative to DeepEval, [RAGAS](https://docs.ragas.io/en/latest/index.html) as a framework is easy to run and straightforward. However, RAGAS is limited to a small set of LLM-only metrics that would require us to use an additional framework on top of to utilize custom metrics. Metrics from RAGAS are implemented directly in DeepEval, so they can still be utilized with DeepEval.
+
+**Arize Phoenix**: [Phoenix](https://phoenix.arize.com/) is another framework focused on both evaluation and observability. Phoenix has a lot to offer in terms of visualizations and LLM tracing, but it's included evaluation tools were a little more limited in comparison to the other frameworks. Considering that the source of your evaluations in Phoenix do not necessarily need to come from Phoenix, if llm tracing is a desired feature for developers or mission heroes, any source of evals can be integrated with this toolset.
+
+## Related ADRs
+This ADR will eventually influence the ADR for overall LLM evaluations.
```
gphorvath, line 54: Might be worth adding any additional context you have here in case it's a while before we start that ADR.

adr/0004-rag-eval-framework.md
```diff
@@ -0,0 +1,59 @@
...
+    - Baseline for performance (tracking how tests perform over time)
+    - Showing to mission heroes (what performance factors do they care about/need to see?)
+  - How are evals tracked/delivered?
+    - How are eval results conveyed?
+    - What internal tools are used to track evals over time?
+
+## Rationale
+The major RAG evaluation frameworks all revolve around LLM-enabled evaluations. While these are fast to get running and offer a lot of flexibility, our use case will likely demand custom evaluation metrics, likely consisting of heuristics testing that do not leverage LLMs for evaluation. Among the frameworks reviewed (see [Alernatives](#alternatives)), DeepEval was found to have the most flexibility in terms of custom evaluation metrics (llm-based and non-llm), as well as what models we want to use for evaluations. DeepEval also offers a toolset for generating evaluation datasets from documentation, which will assist in setting up the datasets used for evaluations. Lastly, DeepEval's test-focused approach will integrate well with the testing workflows in LeapfrogAI.
+
+## Alternatives
```
gphorvath, line 48: Are there other alternatives that we should list, even if we didn't have time or think it necessary to evaluate them?

adr/0004-rag-eval-framework.md
```diff
@@ -0,0 +1,59 @@
...
+  - [Related ADRs](#related-adrs)
+  - [References](#references)
+
+## Status
+
+PROPOSED
+
+## Context
+
+LeapfrogAI uses RAG to provide context-aware responses to users who have specific data they need to reference. In order to make sure RAG is operating at the levels we need it to, we need to get measurable feedback from our RAG pipeline to make it better. We also need a standard to show to mission heroes that we are in fact operating at that level. We do this with RAG-focused evaluations. This ADR Proposes **[DeepEval](https://docs.confident-ai.com/)** as the framework of choice for handling RAG evaluations.
```
gphorvath, line 21: I think this ADR should propose more than a framework. Ideally we would propose a few starting metrics and dataset(s) as well so that the next issue could be the MVP implementation.

adr/0004-rag-eval-framework.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Decision
```
gphorvath, line 23: I would include some starting metrics and open source dataset(s) that we could use in our MVP implementation.

adr/0004-rag-eval-framework.md
```diff
@@ -0,0 +1,59 @@
...
+  - Use those documents to generate questions/ground truth statements using DeepEval tools
+  - Refine those question/ground truth pairs
+- Begin regularly using Evals
+  - Which tests serve what purpose?
+    - e2e testing (e.g deployments are only allowed on passing evals)
+    - Baseline for performance (tracking how tests perform over time)
+    - Showing to mission heroes (what performance factors do they care about/need to see?)
+  - How are evals tracked/delivered?
+    - How are eval results conveyed?
+    - What internal tools are used to track evals over time?
```
gphorvath, line 43: I would remove mission hero specific language here. As far as LeapfrogAI as a product, we should have a generally applicable set of evals, probably derived from open source, as an example for how to properly use this evaluation framework and show/demonstrate the performance of LeapfrogAI against common benchmarks/metrics. Delivery can then create additional custom tests to target mission hero specific use cases.

adr/0004-rag-eval-framework.md
barronstone: IMO, this ADR recommends an evaluation `tool` rather than a `framework`. To me, a `framework` will be more holistic and encompass all of the things listed in the **Decision** section: metrics, dataset, integration into e2e, tracking and conveying results, etc. Deciding on which `tool` we use for evals is an important steps towards that end goal, but I'd backoff on calling this a `framework`.

adr/0004-rag-eval-framework.md
jalling97: Additional context as in what specifically made it feel more limited? Or were you thinking something else?

adr/0004-rag-eval-framework.md
jalling97: A thread came up asking about LangChain, which it may be worthwhile to mention as it technically has an evaluation framework, but each of these can plug into LangChain if/when we want better traceability. Other than that nothing comes to mind for me.

adr/0004-rag-eval-framework.md
barronstone: I think I disagree with @gphorvath (or maybe I'm misunderstanding his comment). If the product team is only using common open-source datasets to evaluate models, how does that differentiate what we're doing from existing model scorecards and things that are already out there using those datasets? I think we need to generate "generally representative of DoD-things" datasets, for example doing RAG with NIST controls, Joint Pubs, DFARS, etc.

I agree with @gphorvath that Delivery can then create additional tests for specific mission heroes' use cases, utilizing actual mission data.

adr/0004-rag-eval-framework.md
barronstone: It would be good to mention Langsmith and why it wasn't included in your evaluation; some of the stuff in your Slack convo with Marshall earlier today

adr/0004-rag-eval-framework.md
jalling97: Based on the couple conversations I've had on the topic, I can definitely add some specific metrics we'll likely care about, but I'm hesitant to include any datasets, mostly because I think in order to reach MVP we'll need to curate a custom dataset from the start. That's just my opinion though and am happy to discuss further.

adr/0004-rag-eval-framework.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Alternatives
+**RAGAS**: A more lightweight alternative to DeepEval, [RAGAS](https://docs.ragas.io/en/latest/index.html) as a framework is easy to run and straightforward. However, RAGAS is limited to a small set of LLM-only metrics that would require us to use an additional framework on top of to utilize custom metrics. Metrics from RAGAS are implemented directly in DeepEval, so they can still be utilized with DeepEval.
+
+**Arize Phoenix**: [Phoenix](https://phoenix.arize.com/) is another framework focused on both evaluation and observability. Phoenix has a lot to offer in terms of visualizations and LLM tracing, but it's included evaluation tools were a little more limited in comparison to the other frameworks. Considering that the source of your evaluations in Phoenix do not necessarily need to come from Phoenix, if llm tracing is a desired feature for developers or mission heroes, any source of evals can be integrated with this toolset.
```
barronstone, line 51: I would clarify that the visualization provided by Arize Phoeniz are very developer-centric. (At least that's how I interpreted them from the demo.) I don't think they will be super-helpful for communicating our eval results/decisions with mission heroes.

adr/0004-rag-eval-framework.md
```diff
@@ -0,0 +1,59 @@
...
+    - e2e testing (e.g deployments are only allowed on passing evals)
+    - Baseline for performance (tracking how tests perform over time)
+    - Showing to mission heroes (what performance factors do they care about/need to see?)
+  - How are evals tracked/delivered?
+    - How are eval results conveyed?
+    - What internal tools are used to track evals over time?
+
+## Rationale
+The major RAG evaluation frameworks all revolve around LLM-enabled evaluations. While these are fast to get running and offer a lot of flexibility, our use case will likely demand custom evaluation metrics, likely consisting of heuristics testing that do not leverage LLMs for evaluation. Among the frameworks reviewed (see [Alernatives](#alternatives)), DeepEval was found to have the most flexibility in terms of custom evaluation metrics (llm-based and non-llm), as well as what models we want to use for evaluations. DeepEval also offers a toolset for generating evaluation datasets from documentation, which will assist in setting up the datasets used for evaluations. Lastly, DeepEval's test-focused approach will integrate well with the testing workflows in LeapfrogAI.
```
barronstone, line 46: The parenthetical statement "llm-based and non-llm" is confusing. Clarify what that means.

barronstone (changes requested):
Good start! Thanks for pulling this together

adr/0004-rag-eval-framework.md
gphorvath: Just any details about how this related to the LLM Eval ADR so we know how it might influence it. A couple thoughts -

1. Did we pick a framework that works for both LLM and RAG?
2. Is there overlap between certain metrics?

adr/0004-rag-eval-framework.md
gphorvath: As Barron suggested above, I think if we pivot this PR to be the RAG Evaluation tool, then we can get away with not raising any datasets. That said, I think we should adopt some open source datasets in our evals in addition to custom dataset so we have some basis of comparison to other literature published by those without access to our data.

adr/0004-rag-eval-framework.md
gphorvath: I agree we will need some custom tests, but having some of the same baseline evals used in open source gives us a basis for comparison against literature on other tools published by those who don't have access to our data.

adr/0004-rag-eval-framework.md
gphorvath: I think the differentiator here is that LLM-based implies an Evaluation LLM is used and non-LLM is more just raw math.

adr/0004-rag-eval-framework.md
jalling97: I added a note about how DeepEval can be used for general LLM evaluations and the section on RAGAS mentions that RAGAS-specific metrics are implemented directly in DeepEval.

adr/0004-rag-eval-framework.md
jalling97: I added a section on LangChain/LangSmith :+1:

adr/0004-rag-eval-framework.md
jalling97: I changed the references from `framework` to `toolset` to help make a better distinction :+1:

adr/0004-rag-eval-framework.md
jalling97: That's fair. There will definitely be a lot of research to be done on a good representative open source dataset (i.e not just using the first one I could find :sweat_smile:), but since I changed references from `framework` to `toolset`, I think that can be a follow-up decision post `0.8.0`. How does that sound?

adr/0004-rag-eval-framework.md
jalling97: Using a baseline open source dataset (TBD which one) certainly doesn't harm us in our evals goals, but I feel like even on the product side we should make a custom dataset that is more representative of the types of tasks we'll need to do RAG on in a delivery context. My high-level looks into dataset options didn't provide much, so I think we'll benefit from a custom dataset approach. And in that case, we can even potentially provide it to HuggingFace.

adr/0004-rag-eval-framework.md
jalling97: I added a statement defining the two directly :+1:

PR #533, opened 2024-05-16
493 add authentication

CollectiveUnicorn:


supabase/migrations/20240419164109_init-assistant.sql
```diff
@@ -2,6 +2,7 @@
 create table
   assistant_objects (
     id uuid primary key DEFAULT uuid_generate_v4(),
+    user_id uuid references auth.users not null,
```
CollectiveUnicorn, line 5: Check the UserID

CollectiveUnicorn: The insert is not working correctly with RLS, this needs to be resolved.

CollectiveUnicorn:
Post requests to create assistant need double checking

supabase/migrations/20240419164109_init-assistant.sql
CollectiveUnicorn: All the endpoints work now with Supabase and auth

src/leapfrogai_api/data/crud_file_object.py
```diff
@@ -1,34 +1,52 @@
...
+        user_id: str = (await self.db.auth.get_user(self.jwt)).user.id
+        return await super().update(
+            id_=id_, object_=AuthFileObject(user_id=user_id, **object_.dict())
+        )
 
-    async def delete(self, id_: str, db: AsyncClient) -> bool:
+    async def delete(self, id_: str) -> bool:
         """Delete a file object by its ID."""
-        return await super().delete(id_=id_, db=db)
+        return await super().delete(id_=id_)
```
gphorvath, line 52: Not sure if I missed something, but this doesn't seem to verify the user before deleting.

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -194,10 +210,13 @@ async def modify_assistant(
...
+    auth_creds: Annotated[HTTPAuthorizationCredentials, Depends(security)],
+) -> AssistantDeleted:
     """Delete an assistant."""
-    crud_assistant = CRUDAssistant(model=Assistant)
-    assistant_deleted = await crud_assistant.delete(db=session, id_=assistant_id)
+    crud_assistant = await CRUDAssistant(auth_creds.credentials)
+    assistant_deleted = await crud_assistant.delete(id_=assistant_id)
     return AssistantDeleted(
         id=assistant_id,
         deleted=bool(assistant_deleted),
```
gphorvath, line 222: Is it possible for the user to delete another user's assistant?

gphorvath (changes requested):

src/leapfrogai_api/README.md
```diff
@@ -41,3 +41,7 @@ From this directory run:
 ``` bash
 make test-integration
 ```
+
+## Notes
+
+* All API calls to `assistants` or `files` endpoints must be authenticated via a Supabase JWT token in the message's `Authorization` header.
```
gphorvath, line 47: How does one setup auth for local dev / swagger?

gphorvath: Actually I see it in the Supabase README. Maybe just a reference to that README here.

.github/workflows/e2e.yaml
```diff
@@ -106,6 +106,17 @@ jobs:
             docker image prune -af
             uds zarf package deploy packages/api/zarf-package-leapfrogai-api-amd64-e2e-test.tar.zst --set=EXPOSE_OPENAPI_SCHEMA=true --confirm
             rm packages/api/zarf-package-leapfrogai-api-amd64-e2e-test.tar.zst
+            
+        - name: Set environment variable
+          id: set-env-var
+          run: |
+            echo "SERVICE_KEY=$(uds zarf tools kubectl get secret supabase-bootstrap-jwt -n leapfrogai -o jsonpath='{.data.service-key}' | base64 -d)" >> "$GITHUB_ENV"
```
gphorvath, line 113: are we still using the service key?

src/leapfrogai_api/data/crud_assistant_object.py
```diff
@@ -1,34 +1,51 @@
...
+            self, jwt=jwt, db=db, model=AuthAssistant, table_name=table_name
+        )
 
-    async def create(self, db: AsyncClient, object_: Assistant) -> Assistant | None:
+    async def create(self, object_: Assistant) -> Assistant | None:
         """Create a new assistant."""
-        return await super().create(db=db, object_=object_)
+        user_id: str = (await self.db.auth.get_user(self.jwt)).user.id
+        return await super().create(
+            object_=AuthAssistant(user_id=user_id, **object_.dict())
```
gphorvath, line 31: object_=AuthAssistant(user_id=user_id, object=object_.model_dump())

.dict() is throwing deprecated.

src/leapfrogai_api/routers/openai/files.py
```diff
@@ -66,20 +70,29 @@ async def list_files(session: Session) -> ListFilesResponse:
...
+    crud_file_object = await CRUDFileObject(auth_creds.credentials)
+    return await crud_file_object.get(id_=file_id)
 
 
 @router.delete("/{file_id}")
-async def delete_file(session: Session, file_id: str) -> FileDeleted:
+async def delete_file(
+    file_id: str,
+    auth_creds: Annotated[HTTPAuthorizationCredentials, Depends(security)],
+) -> FileDeleted:
```
gphorvath, line 86: When I delete a file by ID, the file object gets deleted but not the file from the bucket.

src/leapfrogai_api/routers/openai/files.py
```diff
@@ -66,20 +70,29 @@ async def list_files(session: Session) -> ListFilesResponse:
...
 
-    crud_file_bucket = CRUDFileBucket(model=UploadFile)
-    await crud_file_bucket.delete(client=session, id_=file_id)
+    crud_file_object = await CRUDFileObject(auth_creds.credentials)
+    file_deleted = await crud_file_object.delete(id_=file_id)
+
+    crud_file_bucket = await CRUDFileBucket(
+        jwt=auth_creds.credentials, model=UploadFile
+    )
+    await crud_file_bucket.delete(id_=file_id)
```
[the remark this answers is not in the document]
gphorvath: Generally seems like this is a non-issue

src/leapfrogai_api/data/crud_assistant_object.py
gphorvath: this is done in a few places

src/leapfrogai_api/README.md
CollectiveUnicorn: Added some more information to the README to make it more clear.

src/leapfrogai_api/routers/openai/files.py
CollectiveUnicorn: This has been corrected with RLS and python updates that check if the delete was successful on the table prior to trying to delete from the bucket.

src/leapfrogai_api/data/crud_assistant_object.py
CollectiveUnicorn: I've replace all the dict references with model_dump()

tests/e2e/test_api.py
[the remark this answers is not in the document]
CollectiveUnicorn: Will check on the tests, going to explicitly call out the tests that I added as auth tests.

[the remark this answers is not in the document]
CollectiveUnicorn: I have updated the integration test to successfully auth when the environment variable SUPABSE_USER_JWT is present. So all the integration tests should be working again.

src/leapfrogai_api/routers/supabase_session.py
```diff
@@ -1,17 +1,74 @@
...
+from fastapi import Depends, status, HTTPException
 from supabase_py_async import AsyncClient, create_client
+from httpx import HTTPStatusError
+from gotrue import errors, types
 
 
 async def init_supabase_client() -> AsyncClient:
     """Initialize a Supabase client."""
+    logging.info(
+        f"The anon key is {os.getenv('SUPABASE_ANON_KEY')} the url is {os.getenv('SUPABASE_URL')}"
```
YrrepNoj, line 15: Should we be logging this?

src/leapfrogai_api/routers/supabase_session.py
```diff
@@ -1,17 +1,74 @@
...
+        session (Session): the default anonymous session
+        authorization (str): the JWT token for the user
+    """
+
+    authorized = True
+
+    if authorization:
+        api_key: str = authorization.replace("Bearer ", "")
+
+        logging.info(f"The api key is {api_key}")
```
YrrepNoj, line 40: Should we be logging the api key? That sounds like a potential security concern.

src/leapfrogai_api/main.py
```diff
@@ -35,6 +40,26 @@ async def lifespan(app: FastAPI):
...
+
+skip_endpoint_auth = ["/healthz", "/docs", "/openapi.json"]
+
+
+@app.middleware("http")
+async def verify_supabase_auth(request: Request, call_next):
+    if request.url.path in skip_endpoint_auth:
+        return await call_next(request)
+    else:
+        anon_session = await init_supabase_client()
```
YrrepNoj, line 52: Every http request is going to spawn a new client to supabase? That seems like something that will start to become painful when concurrent requests are happening.

Does it make sense to create a client (or pool of clients) at startup and use those instead of creating a new session during every request?

src/leapfrogai_api/main.py
```diff
@@ -35,6 +40,26 @@ async def lifespan(app: FastAPI):
...
+        authorization_header = request.headers.get("authorization")
+
+        try:
+            await validate_user_authorization(anon_session, authorization_header)
+        except HTTPException as e:
+            return JSONResponse(status_code=e.status_code, content={})
```
YrrepNoj, line 58: Can we return more than just a status code? Why not bubble down the exception description?

src/leapfrogai_api/data/crud_assistant_object.py
```diff
@@ -1,34 +1,51 @@
 """CRUD Operations for Assistant."""
 
+from pydantic import Field
 from openai.types.beta import Assistant
 from supabase_py_async import AsyncClient
```
YrrepNoj, line 5: Is there a particular we are using the `supabase_py_async` library? I searched our current codebase and anytime we use that module we are `await`'ing the result. 

This async library seems to be complicating our repo by having us cater to the async classes such as defining these new `__ainit__()` in several places. 

Having an async library could be cool, but it feels like we are misusing it. We are not getting any of the benefits of having async classes/functions, but we are still paying the tech-complication price of trying to do async things with Python.

src/leapfrogai_api/README.md
```diff
@@ -8,22 +8,34 @@ A mostly OpenAI compliant API surface.
...
+    ``` bash
+    supabase start # from /leapfrogai
+    
+    supabase db reset # clears all data and reinitializes migrations
+    
+    supabase status # to check status and see your keys
+    ```
 
-supabase db reset # clears all data and reinitializes migrations
+2. Go to your local Supabase dashboard and create a test user.
```
YrrepNoj, line 21: This step is a bit ambiguous, can you give more instructions on what a 'test user' is and how one could create it?

The default supabase configuration doesn't have user/password auth right? Does a config need to change in order to complete this step?

YrrepNoj (changes requested):

src/leapfrogai_api/README.md
CollectiveUnicorn: Should be a bit better now

src/leapfrogai_api/main.py
CollectiveUnicorn: Passed along the detail into the json response

src/leapfrogai_api/main.py
CollectiveUnicorn: I have replaced the call to `anon_session = await init_supabase_client()` with a single instance that is used for all validations, in this case I don't think there is a need for a pool of clients. 

Separate from this, there will however still be new clients created when the user hits an endpoint that uses Supabase, due to needing to have a client that has authenticated using that user's specific JWT.

src/leapfrogai_api/data/async_mixin.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, *args, **kwargs):
+        """
+        Standard constructor used for arguments pass
+        Do not override. Use __ainit__ instead
+        """
+        self.__storedargs = args, kwargs
+        self.async_initialized = False
+
+    async def __ainit__(self, *args, **kwargs):
+        """Async constructor, you should implement this"""
```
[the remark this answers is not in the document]
CollectiveUnicorn: We don't, I've added a pass and updated the comment to be more explicit about what needs to be done.

src/leapfrogai_api/routers/supabase_session.py
CollectiveUnicorn: Nope, removed

src/leapfrogai_api/routers/supabase_session.py
CollectiveUnicorn: Nope, removed

src/leapfrogai_api/data/crud_assistant_object.py
CollectiveUnicorn: We can do the vast majority of things without async I think. Though the addition of async was a refactor that came out of db concurrency concerns. Fast api will run each endpoint call in a separate thread so as long as there's not shared objects between those threads it should be fine. I'll touch base with you for more context though and determine if a refactor is warranted in this PR.

CollectiveUnicorn: Ended up moving everything into the `supabase_session.init_supabase_client` call getting the JWT token and setting up the user session via some `depends` chaining with a dummy refresh token. Afterwhich as long as `session: Session` is included in the api's function signature then everything will be authed and validated for the user.

CollectiveUnicorn:
Issue for long-lived API keys can be found here: https://github.com/defenseunicorns/leapfrogai/issues/561

src/leapfrogai_api/README.md
alekst23: @YrrepNoj I have added better instructions and a makefile script to do this

src/leapfrogai_api/main.py
alekst23: Based on the current implementation, I believe the client will be instantiated once per session. However, it should also be scoped to the session and be released once the call returns.

alekst23 (dismissed):
I think this is a great start and very cool way to implement Auth dependencies.

Our testing will require some work, but I would like to address this in it's own PR.

src/leapfrogai_api/README.md
```diff
@@ -2,28 +2,44 @@
...
+2. Create a local Supabase instance (requires [Supabase CLI](https://supabase.com/docs/guides/cli/getting-started)):
+    ```bash
+    brew install supabase/tap/supabases
 
-``` bash
-supabase start # from /leapfrogai
+    supabase start # from /leapfrogai
 
-supabase db reset # clears all data and reinitializes migrations
+    supabase stop --project-id leapfrogai_api # stop api containers
```
gphorvath, line 20: Is this the right project-id? For my instance it's just `leapfrogai`

alekst23: Seems like both work, but `leapfrogai` is probably the more proper form.

supabase/migrations/20240419164109_init-assistant.sql
```diff
@@ -2,6 +2,7 @@
...
     created_at bigint default extract(epoch from now()) not null,
```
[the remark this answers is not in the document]
gphorvath: done

tests/integration/api/test_files.py
```diff
@@ -10,14 +12,31 @@
...
+client = TestClient(router, headers=headers)
 
 
 @pytest.fixture(scope="session", autouse=True)
 def read_testfile():
     """Read the test file content."""
     global testfile_content  # pylint: disable=global-statement
-    with open("tests/data/test.txt", "rb") as testfile:
+    with open(
+        os.path.dirname(__file__) + "/../../../tests/data/test.txt", "rb"
```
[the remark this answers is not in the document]
gphorvath: Good catch, fixed.

tests/pytest/leapfrogai_api/test_api.py
```diff
@@ -21,7 +30,48 @@
...
+
+
+async def mock_init_supabase_client(
+    auth_creds: Annotated[HTTPAuthorizationCredentials, Depends(security)],
+) -> AsyncClient:
+    return AsyncClient("", "", "", ClientOptions())
+
+
+async def pack_dummy_bearer_token(request: _CachedRequest, call_next):
+    request.headers.__dict__["_list"].append(
```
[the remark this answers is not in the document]
gphorvath: yes

gphorvath (dismissed):

gphorvath (approved):

alekst23 (approved):

YrrepNoj (approved):

Issue #418, opened 2024-04-18
feat(api): Messages Endpoints


gphorvath:
### User Story

**As a** LeapfrogAI Developer 
**I want** an OpenAI compliant Messages API
**So that** I can exchange messages via the API

### Acceptance Criteria

Implement and test the following endpoints:
- [ ] Create Message
- [ ] List Messages
- [ ] Retrieve Message
- [ ] Modify Message

### Additional context
- https://platform.openai.com/docs/api-reference/messages

Issue #417, opened 2024-04-18
feat(api): Threads Endpoints


gphorvath:
### User Story

**As a** LeapfrogAI Developer 
**I want** an OpenAI compliant Threads API
**So that** I can store message history via the API.

### Acceptance Criteria

Implement and test the following endpoints:
- [ ] Create thread
- [ ] Retrieve Thread
- [ ] Modify Thread
- [ ] Delete Thread

### Additional context
- https://platform.openai.com/docs/api-reference/threads

PR #554, opened 2024-05-22
feat(api): Threads and Message Endpoints

CollectiveUnicorn:
- Updates OpenAI library to 1.30.5
- Implements the thread endpoints that match [OpenAI message spec](https://platform.openai.com/docs/api-reference/messages)
- Implements the message endpoints that match the [OpenAI threads spec](https://platform.openai.com/docs/api-reference/threads)
- Creates DB schema for threads, messages, runs, and run steps

Closes: https://github.com/defenseunicorns/leapfrogai/issues/417
Closes: https://github.com/defenseunicorns/leapfrogai/issues/418


src/leapfrogai_api/backend/types.py
```diff
@@ -365,6 +367,44 @@ class ListVectorStoresResponse(BaseModel):
...
+class CreateMessageRequest(BaseModel):
+    """Request object for creating a message."""
+
+    role: Literal["user", "assistant"] = Field(default="user")
+    content: list[MessageContent] = Field(
+        default=[TextContentBlock(text=Text(value="", annotations=[]), type="text")],
+        examples=[[TextContentBlock(text=Text(value="", annotations=[]), type="text")]],
+    )
+    attachments: Optional[List[Attachment]] = Field(default=None, examples=[None])
+    metadata: Optional[dict] = Field(default=None)
```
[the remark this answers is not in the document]
CollectiveUnicorn: I had these initially as object but was running into issues with FastAPI serialization when trying to use object instead of dict. Is vscode suggesting using Optional[object]?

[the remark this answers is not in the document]
CollectiveUnicorn: But some issues may have gotten conflated together, I'll try the object again.

src/leapfrogai_api/routers/openai/threads.py
```diff
@@ -1,74 +1,240 @@
...
+
+        thread = Thread(
+            id="",  # Leave blank to have Postgres generate a UUID
+            created_at=0,  # Leave blank to have Postgres generate a timestamp
+            metadata=request.metadata,
+            object="thread",
+            tool_resources=request.tool_resources,
+        )
+
+        new_thread: Thread = await crud_thread.create(object_=thread)
```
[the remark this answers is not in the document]
CollectiveUnicorn: Fixed this

src/leapfrogai_api/.gitignore
gphorvath, line 1: I have another comment below on the .jwt file. Depending on how you resolve that you may not need this .gitignore. 

If you do need it, I would move this to the root .gitignore so we don't accidently push the .jwt file if it gets generated anywhere in the project hierarchy.

src/leapfrogai_api/Makefile
```diff
@@ -23,16 +20,23 @@ define get_jwt_token
...
-	echo "export SUPABASE_USER_JWT=$$JWT" > .jwt; \
+	echo "SUPABASE_USER_JWT=$$JWT" > .jwt; \
+	echo "SUPABASE_URL=$$SUPABASE_URL" >> .jwt; \
+	echo "SUPABASE_ANON_KEY=$$SUPABASE_ANON_KEY" >> .jwt; \
 	echo "DONE - JWT token copied to clipboard"
 endef
 
-supabase-user:
-	@read -s -p "Enter your Supabase password: " SUPABASE_PASS; echo; \
+new-api-user:
```
gphorvath, line 29: nit - maybe just call this `user` instead of `new-api-user`

src/leapfrogai_api/backend/types.py
```diff
@@ -4,13 +4,15 @@
 
 import datetime
 from enum import Enum
-from typing import Literal
+from typing import Literal, Optional, List
```
gphorvath, line 7: Minor nit, but In a previous PR we had discussed switch to `list` and `| None` instead of `List` and `Optional`. It's a little bit of a break from the convention of the rest of this file to bring them back.

src/leapfrogai_api/data/crud_run.py
```diff
@@ -0,0 +1,55 @@
...
+    user_id: str = Field(default="")
+
+
+class CRUDRun(CRUDBase[AuthRun]):
+    """CRUD Operations for run"""
+
+    def __init__(self, db: AsyncClient):
+        super().__init__(db=db, model=AuthRun, table_name="run_objects")
+
+    async def create(self, object_: Run) -> AuthRun | None:
```
gphorvath, line 21: I think we decided to switch to return the base object instead of the Auth object. It looks like you return the base object for Threads, but the Auth object for runs/messages. Any reason for the delta?

supabase/migrations/20240522141100_threads_runs_messages_steps.sql
```diff
@@ -0,0 +1,177 @@
...
+-- Foreign key constraints for run_step_objects
+alter table run_step_objects add constraint fk_run_step_objects_thread_objects
+    foreign key (thread_id)
+    REFERENCES thread_objects (id);
+alter table run_step_objects add constraint fk_run_step_objects_assistant_objects
+    foreign key (assistant_id)
+    REFERENCES assistant_objects (id);
+alter table run_step_objects add constraint fk_run_step_objects_run_objects
+    foreign key (run_id)
+    REFERENCES run_objects (id);
```
gphorvath, line 111: Minor nit, but can't these be declared on table creation instead of altering the table since we are adding all these in the same PR?

src/leapfrogai_api/Makefile
```diff
@@ -23,16 +20,23 @@ define get_jwt_token
 	echo "Extracting token from $(TOKEN_RESPONSE)"; \
 	JWT=$$(echo $$TOKEN_RESPONSE | grep -oP '(?<="access_token":")[^"]*'); \
 	echo -n "$$JWT" | xclip -selection clipboard; \
-	echo "export SUPABASE_USER_JWT=$$JWT" > .jwt; \
+	echo "SUPABASE_USER_JWT=$$JWT" > .jwt; \
+	echo "SUPABASE_URL=$$SUPABASE_URL" >> .jwt; \
+	echo "SUPABASE_ANON_KEY=$$SUPABASE_ANON_KEY" >> .jwt; \
```
gphorvath, line 25: I've actually been thinking about this one... The testing feature in VS Code automatically hooks into the `.env` file in the root of the project and we already gitignore .env... maybe we should be writing to that instead?

src/leapfrogai_api/backend/types.py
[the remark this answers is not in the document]
gphorvath: We have this issue in several places object vs dict on the types...we probably should test with an OpenAI client library and update all the types to be consistent.

gphorvath (changes requested):

supabase/migrations/20240522141100_threads_runs_messages_steps.sql
gphorvath, line 1: this needs to move to `leapfrogai/packages/api/supabase/migrations`

src/leapfrogai_api/data/crud_run.py
gphorvath: resolved

supabase/migrations/20240522141100_threads_runs_messages_steps.sql
gphorvath: moved

supabase/migrations/20240522141100_threads_runs_messages_steps.sql
gphorvath: done

src/leapfrogai_api/backend/types.py
gphorvath: resolved to be consistent with the rest of types.

src/leapfrogai_api/Makefile
gphorvath: done

src/leapfrogai_api/Makefile
gphorvath: done

src/leapfrogai_api/.gitignore
gphorvath: done

gphorvath (dismissed):

tests/integration/api/test_vector_stores.py
```diff
@@ -178,7 +178,6 @@ def test_modify(create_file):
     request = ModifyVectorStoreRequest(
         file_ids=[create_file["id"]],
         name="test1",
-        expires_after=ExpiresAfter(anchor="last_active_at", days=0),
```
YrrepNoj, line 181: Question: why not set the `expires_after` field anymore?

gphorvath: In this particular case we were encounter an issue where the vector store expires before the next step in the test. We also test for expiration still, so this test doesn't need to expire.

gphorvath (approved):

YrrepNoj (approved):

PR #570, opened 2024-05-29
chore: better database migrations - managed by supabase

YrrepNoj:
This PR introduces a new approach to how we perform data migrations for our LFAI applications.

1) Instead of running the migration scripts as part of the entrypoint scripts of the POSTGRES containers we are running the migrations as a [k8s job](https://kubernetes.io/docs/concepts/workloads/controllers/job/) for each deployment that needs the migration (in this case the API and UI deployments).
2) Instead of running the migration scripts with raw PSQL we are utilizing the [supabase cli](https://supabase.com/docs/guides/cli/getting-started?queryGroups=platform&platform=linux) to perform and manage our migrations.
  -  The supabase CLI helps manage our migrations by tracking the history of past migration scripts that got run. With the CLI rerunning the same migration will not break as it will know the migration script already ran and won't attempt to re-run it.


jalling97 (dismissed):

packages/api/supabase/migrations/20240422015807_v0.7.0-all.sql
```diff
@@ -10,9 +10,3 @@ create table
     status text,
     status_details text
   );
-
--- storage bucket for the files
-insert into storage.buckets
-  (id, name, public)
-values
-  ('file_bucket', 'files', true);
```
CollectiveUnicorn, line 18: The removal of this code causes the openai files endpoint's upload to fail. As there's no longer an entry in `storage.buckets` for the `files` folder, so it effectively does not exist. There is a query [here](https://github.com/defenseunicorns/leapfrogai/blob/f31f1905f96e5fcaf10cacca88f931384e59a8bc/packages/api/supabase/migrations/20240516152530_enable_rls.sql#L7) that sets it to private, you can probably set it to private within the insert command instead without issue.

CollectiveUnicorn (changes requested):

packages/api/supabase/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM cgr.dev/chainguard/bash:latest
+ARG version="1.169.8"
+ENV version=${version}
+
+# Download the supabase cli
+RUN mkdir /usr/local/bin
+RUN curl -LO https://github.com/supabase/cli/releases/download/v${version}/supabase_linux_amd64.tar.gz
+RUN tar -xzf supabase_linux_amd64.tar.gz -C /usr/local/bin/
+RUN rm supabase_linux_amd64.tar.gz
```
CollectiveUnicorn, line 9: Do we need to grab arm releases as well?

src/leapfrogai_ui/supabase/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM cgr.dev/chainguard/bash:latest
+ARG version="1.169.8"
+ENV version=${version}
+
+# Download the supabase cli
+RUN mkdir /usr/local/bin
+RUN curl -LO https://github.com/supabase/cli/releases/download/v${version}/supabase_linux_amd64.tar.gz
+RUN tar -xzf supabase_linux_amd64.tar.gz -C /usr/local/bin/
+RUN rm supabase_linux_amd64.tar.gz
```
CollectiveUnicorn, line 9: Do we need to grab arm releases as well?

CollectiveUnicorn (changes requested):

packages/api/supabase/migrations/20240422015807_v0.7.0-all.sql
YrrepNoj: good catch, added

src/leapfrogai_ui/supabase/Dockerfile
YrrepNoj: Yes, good catch. Added logic so this works for arm targets as well

packages/api/supabase/Dockerfile
YrrepNoj: Yes, good catch. Added logic so this works for arm targets as well

CollectiveUnicorn (approved):

jalling97 (approved):

Issue #419, opened 2024-04-18
feat(api): Runs endpoints


gphorvath:
### User Story

**As a** LeapfrogAI Developer 
**I want** an OpenAI compliant Runs API
**So that** I can execute actions on threads via the API.

### Acceptance Criteria

Implement and test the following endpoints:
- [ ] Create run
- [ ] Create thread and run
- [ ] List runs
- [ ] Retrieve run
- [ ] Modify run
- [ ] Submit tool outputs to run
- [ ] Cancel run

### Additional context
- https://platform.openai.com/docs/api-reference/runs

PR #583, opened 2024-06-03
feat(api): Runs endpoints

CollectiveUnicorn:
Streaming and non-streaming versions of runs
Naive RAG implementation that allows generating responses using results pulled from the vector store

Closes https://github.com/defenseunicorns/leapfrogai/issues/419


src/leapfrogai_api/backend/converters.py
```diff
@@ -0,0 +1,111 @@
...
+    TextDeltaBlock,
+    TextDelta,
+)
+from pydantic_core import ValidationError
+
+from leapfrogai_api.backend.validators import TextContentBlockParamValidator
+from leapfrogai_api.routers.openai.requests.create_message_request import (
+    CreateMessageRequest,
+)
+from leapfrogai_api.routers.openai.threads import create_message
```
gphorvath, line 24: We should avoid an import from the router. This is causing a circular import. We should pull this into it's own class, similarly to index.py, and have both the router and this take a dependency on it.

src/leapfrogai_api/routers/openai/requests/create_thread_request.py
```diff
@@ -0,0 +1,60 @@
...
+from openai.types.beta import Thread
+from openai.types.beta.thread import ToolResources as BetaThreadToolResources
+from openai.types.beta.threads import Message
+from pydantic import BaseModel, Field
+
+from leapfrogai_api.routers.openai.requests.create_message_request import (
+    CreateMessageRequest,
+)
+from leapfrogai_api.data.crud_thread import CRUDThread
+from leapfrogai_api.routers.openai.threads import create_message, delete_message
```
gphorvath, line 12: We should avoid an import from the router. This is causing a circular import. We should pull this into it's own class, similarly to index.py, and have both the router and this take a dependency on it.

src/leapfrogai_api/routers/openai/requests/run_create_params_request.py
```diff
@@ -0,0 +1,116 @@
...
+    RunCreateParamsRequestBase,
+)
+from leapfrogai_api.routers.openai.requests.create_message_request import (
+    CreateMessageRequest,
+)
+from leapfrogai_api.data.crud_run import CRUDRun
+from leapfrogai_api.backend.converters import (
+    from_content_param_to_content,
+)
+from leapfrogai_api.routers.openai.threads import create_message
```
gphorvath, line 24: We should avoid an import from the router. This is causing a circular import. We should pull this into it's own class, similarly to index.py, and have both the router and this take a dependency on it.

gphorvath (changes requested):
We have a circular import because we are depending on our own route.

make-supabase.mk
```diff
@@ -0,0 +1,29 @@
...
+	echo "SUPABASE_ANON_KEY=$$SUPABASE_ANON_KEY" >> .jwt; \
+	echo "DONE - JWT token copied to clipboard"
+endef
+
+supabase-new-user:
+	@read -s -p "Enter a new DEV API password: " SUPABASE_PASS; echo; \
+	echo "Creating new supabase user..."; \
+	$(call get_jwt_token,"${SUPABASE_URL}/auth/v1/signup")
+
+jwt:
```
YrrepNoj, line 27: Would it make sense to take advantage of the scriptyness of Python and call something like [this utils function](https://github.com/defenseunicorns/leapfrogai/blob/74911a8a2edeafa82488531e382b1b8c3c9c9bd6/tests/e2e/utils.py#L38-L54) here?

This would have the double benefit of
1) testing our test utils even more
2) have only one overhead function of fetching jwts to maintain instead of two.
3) Not have to install additional dependencies (`xclip` used above is not native to all machines)

tests/conftest.py
```diff
@@ -0,0 +1,14 @@
+import os
+from dotenv import load_dotenv
+
+def pytest_configure(config):
```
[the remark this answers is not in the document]
CollectiveUnicorn: This has been reverted in expectation of more comprehensive tests coming in another PR.

src/leapfrogai_api/routers/openai/threads.py
```diff
@@ -25,58 +39,173 @@
...
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Unable to create run",
+        ) from exc
+
 
+@router.post("/runs", response_model=None)
+async def create_thread_and_run(
+    session: Session, request: ThreadRunCreateParamsRequestBaseRequest
+) -> Run | StreamingResponse:
```
gphorvath, line 89: The returned non-streaming Run object contains a user_id.

src/leapfrogai_api/routers/openai/threads.py
```diff
@@ -25,58 +39,173 @@
...
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Unable to create thread",
+        ) from exc
+
+
+@router.post("/{thread_id}/runs", response_model=None)
+async def create_run(
+    thread_id: str, session: Session, request: RunCreateParamsRequestBaseRequest
+) -> Run | StreamingResponse:
```
gphorvath, line 63: The returned non-streaming Run object contains a user_id.

packages/api/supabase/migrations/20240611111500_runs_steps.sql
```diff
@@ -0,0 +1,90 @@
+-- Create a table to store the OpenAI Run Objects
+create table
+  run_objects (
+    id uuid primary key default uuid_generate_v4(),
+    user_id uuid references auth.users not null,
+    object text check (object in ('thread.run')),
+    created_at bigint default extract(epoch FROM NOW()) not null,
+    thread_id uuid references thread_objects (id) on delete cascade ,
+    assistant_id uuid references assistant_objects (id) on delete set null,
```
gphorvath, line 9: I'm starting to think this should be a `cascade` instead of a `set null` so that Runs get cleaned up when an assistant is deleted.

packages/api/zarf.yaml
```diff
@@ -22,6 +22,9 @@ variables:
     default: "uds.dev"
   - name: DEFAULT_EMBEDDINGS_MODEL
     default: "text-embeddings"
+  - name: SUPABASE_CLUSTER_URL
+    description: The URL for Supabase for requests inside the cluster
+    default: http://supabase-kong.leapfrogai.svc.cluster.local:80
```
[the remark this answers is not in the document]
CollectiveUnicorn: Will hardcode this in the values file

src/leapfrogai_api/routers/openai/requests/create_thread_request.py
```diff
@@ -0,0 +1,94 @@
...
+                    thread_id=new_thread.id, message_id=message.id, session=session
+                )
+            raise exc
+
+    async def create_message(self, thread_id: str, session: Session) -> Message:
+        """Create a message."""
+        try:
+            crud_message = CRUDMessage(db=session)
+
+            message_content = await self.get_message_content()
```
CollectiveUnicorn, line 60: Issue with this not working

tests/integration/api/test_threads.py
```diff
@@ -8,11 +8,15 @@
...
 from leapfrogai_api.backend.types import (
-    CreateThreadRequest,
     ModifyThreadRequest,
     ModifyMessageRequest,
+)
+from leapfrogai_api.routers.openai.requests.create_message_request import (
     CreateMessageRequest,
 )
+from leapfrogai_api.routers.openai.requests.create_thread_request import (
+    CreateThreadRequest,
```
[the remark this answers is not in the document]
CollectiveUnicorn: Open for a future refactor will add an issue for this

src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py
```diff
@@ -0,0 +1,376 @@
...
+        return [
+            from_assistant_stream_event_to_str(
+                ThreadRunCompleted(data=run, event="thread.run.completed")
+            )
+        ]
+
+    async def update_with_assistant_data(self, session: Session) -> Assistant:
+        assistant: Assistant | None = await retrieve_assistant(
+            session=session, assistant_id=self.assistant_id
+        )
```
alekst23, line 116:
```suggestion
        crud_assistant = CRUDAssistant(session)
        assistant: Assistant | None = await crud_assistant.get(filters=FilterAssistant(id=self.assistant_id))
```
You will also need to add this to top:
`from leapfrogai_api.data.crud_assistant import CRUDAssistant, FilterAssistant`

src/leapfrogai_api/routers/openai/threads.py
CollectiveUnicorn: Will eliminate

src/leapfrogai_api/routers/openai/requests/create_thread_request.py
CollectiveUnicorn: This has been resolved

src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py
CollectiveUnicorn: This has been resolved

src/leapfrogai_api/routers/openai/threads.py
CollectiveUnicorn: Done

src/leapfrogai_api/routers/openai/threads.py
CollectiveUnicorn: This is has been removed at the runs level

tests/integration/api/test_threads.py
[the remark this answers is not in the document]
CollectiveUnicorn: Issue has been added here: https://github.com/defenseunicorns/leapfrogai/issues/633

make-supabase.mk
CollectiveUnicorn: This has been reverted to be included in more comprehensive tests coming in another PR.

packages/api/supabase/migrations/20240611111500_runs_steps.sql
gphorvath: In testing I determined that assistant_id is a required parameter for the Run object to pass schema validation. I think for now, we should cascade delete to prevent an invalid schema.

gphorvath: fixed

gphorvath (dismissed):

gphorvath:
Latest Updates:
* Removed all run-steps references as we didn't full implement (schema, crud operations, etc.)
* Split up integration tests by route
* Split up threads route into threads/messages/runs
* Removed RAG deployment from UDS bundles (this PR adds it via API)
* Added check for supported tools to threads/assistants endpoints that accept tools and tests to verify.
* Fix: Bug when assistant is deleted and Run remains behind - now run cascade deletes on assistant delete.
* Fix: Modifying run was failing schema check.
* Refactored threads/runs modify endpoints to ensure we aren't overwriting any unmodifiable content.

src/leapfrogai_api/backend/grpc_client.py
```diff
@@ -66,6 +69,22 @@ async def stream_chat_completion(model: Model, request: lfai.ChatCompletionReque
         return StreamingResponse(recv_chat(stream), media_type="text/event-stream")
 
 
+async def stream_chat_completion_raw(
```
YrrepNoj, line 72: Question; when would we call `stream_chat_completion_raw()` vs `stream_chat_completion()`

src/leapfrogai_api/data/crud_run.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, db: AsyncClient):
+        super().__init__(db=db, model=AuthRun, table_name="run_objects")
+
+    async def create(self, object_: Run) -> Run | None:
+        """Create new run."""
+        user_id: str = (await self.db.auth.get_user()).user.id
+        run: AuthRun = await super().create(
+            object_=AuthRun(user_id=user_id, **object_.model_dump())
+        )
+        del run.user_id
```
YrrepNoj, line 27: We are type hinting that this function is returning a `Run obj`. This `run` variable is technically an `AuthRun`. Deleting the `run.user_id` gives. the RunAuth the same 'shape' as a Run, but it is not a Run.

We should probably make sure we are returning an actual Run type instead of something that has the same 'shape' as a Run.

packages/api/supabase/migrations/20240611111500_runs.sql
```diff
@@ -0,0 +1,52 @@
...
+    cancelled_at bigint,
+    failed_at bigint,
+    completed_at bigint,
+    model text,
+    instructions text,
+    tools jsonb,
+    metadata jsonb,
+    parallel_tool_calls boolean,
+    stream boolean,
+    file_ids uuid[],
```
gphorvath, line 24: Should this be a foreign key?

src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py
```diff
@@ -0,0 +1,423 @@
...
+            )
+        )
+        yield "\n\n"
+
+        # The accumulated streaming response
+        response: str = ""
+
+        index: int = 0
+        async for streaming_response in chat_response:
+            random_uuid: UUID = uuid.uuid4()
```
gphorvath, line 390: We probably shouldn't be generating UUIDs in code

YrrepNoj (approved):
This is a big PR that has been the work of a lot of effort. There are several conversation items that will turn into issues in our backlog. Thanks everyone for their hard work in implementing this feature.

In the future, lets set the goal of maintaining smaller PRs and introducing more testing. I'm sure I could spend another couple working days looking through this PR but I believe it would be more efficient to merge this in now (since it is in a working state where we can iterate on it independently) and start working on adding improvements in separate PRs.

Thanks again for everyones hard work in developing, reviewing, and testing this PR.

jalling97 (approved):

packages/api/supabase/migrations/20240611111500_runs.sql
gphorvath: Added ticket for this https://github.com/defenseunicorns/leapfrogai/issues/646

src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py
gphorvath: Added an issue for this: https://github.com/defenseunicorns/leapfrogai/issues/647

src/leapfrogai_api/data/crud_run.py
gphorvath: Added an issue for this: https://github.com/defenseunicorns/leapfrogai/issues/648

PR #586, opened 2024-06-03
chore(test): Create Testing Strategy ADR

alekst23:
## Background
We need to have a consolidated and clearly defined approach to testing our systems. Currently, we don't have a clear definition of what features and functions should be test at each level, and we mix some integration and system testing.

It is also currently difficult for new developers to spin up a test suite in a new environment. We should have tests work out of the box.

## Description
This PR creates an ADR to help us organize our ideas about testing.

## TODO
This file lacks a unified UI testing strategy


adr/0005-testing-strategy
CollectiveUnicorn, line 1: Its missing the `.md` extension

CollectiveUnicorn (changes requested):

adr/0005-testing-strategy
```diff
@@ -0,0 +1,102 @@
...
+  - [Alternatives](#alternatives)
+  - [References](#references)
+  - [Notes](#Notes)
+
+
+## Status 
+- PROPOSED
+
+## Context
+We need to have a consolidated and clearly defined approach to testing our systems. Currently, we don't have a clear definition of what features and functions should be test at each level, and we mix some integration and system testing.
```
CollectiveUnicorn:
```suggestion
We need to have a consolidated and clearly defined approach to testing our systems. Currently, we don't have a clear definition of what features and functions should be tested at each level, and we mix some integration and system testing.
```

adr/0005-testing-strategy
```diff
@@ -0,0 +1,102 @@
...
+
+(3) Tests should be granular, and each test should test an individual feature. Avoid monolithic test functions.
+
+(4) We should be testing for both success and failure cases.
+
+(5) We should use `unittest.mock.patch` to mock features for testing.
+
+(6) We should use `pytest.mark.parametrize` to iterate over parameter lists for test function.
+
+(7) We should use `conftest.py` to set up the dev environment variables, and to create global test fixtures.
```
CollectiveUnicorn: Is it necessary to call out specific library functions for tests?

adr/0005-testing-strategy
```diff
@@ -0,0 +1,102 @@
...
+- PROPOSED
+
+## Context
+We need to have a consolidated and clearly defined approach to testing our systems. Currently, we don't have a clear definition of what features and functions should be test at each level, and we mix some integration and system testing.
+
+It is also currently difficult for new developers to spin up a test suite in a new environment. We should have tests work out of the box.
+
+## Decisions
+
+(1) We should clearly define the levels of testing that we want to do, and the featuers that we should test at each one.
```
jalling97, line 26: misspelling of features

CollectiveUnicorn (changes requested):

adr/0005-testing-strategy.md
```diff
@@ -0,0 +1,102 @@
...
+**PROS**: 
+Faster development, less redundant work.
+Faster and leaner test runs.
+
+**CONS**:
+Coverage is up to devs and can lapse.
+
+(3) Tests should be granular, and each test should test an individual feature. Avoid monolithic test functions.
+
+(4) We should be testing for both success and failure cases.
```
CollectiveUnicorn, line 43: Can we add more detail here, I wouldn't think that its necessary to enumerate all failure cases for instance. But we may need to have a wider discussion about this because I'm sure there's a number of varying perspectives.

adr/0005-testing-strategy.md
```diff
@@ -0,0 +1,102 @@
...
+
+(8) We should consider all testing environments:
+* Windows
+* MacOS
+* Linux
+* Kubernetes
+
+## Design
+
+### Unit Testing
```
CollectiveUnicorn, line 59: I believe there was some desire to minimize the amount of unit tests created to avoid the code becoming overly rigid/brittle. Might be worth adding as discussion topic as a team.

adr/0005-testing-strategy.md
```diff
@@ -0,0 +1,102 @@
...
+
+## References
+* [Understanding the Python Mock Object Library](https://realpython.com/python-mock-library/)
+
+
+* [How to parametrize fixtures and test functions](https://docs.pytest.org/en/stable/how-to/parametrize.html)
+
+## NOTES
+
+We need to integrate UI testing strategy into this document.
```
CollectiveUnicorn, line 102: Will probably need @andrewrisse to really fill this section out, but the UI tests use Playwright atm and cover at least some e2e and integration tests. The svelte server api should probably have all those endpoints tested as well as all the UI features.

CollectiveUnicorn (changes requested):

adr/0005-testing-strategy.md
CollectiveUnicorn: Actually, I'm going to resolve this. Since its a todo.

adr/0005-testing-strategy
alekst23: I did not want to get too granular, but the `conftest.py` file is specific to implementing `pytest` properly so I wanted to point that out for devs.

adr/0005-testing-strategy.md
alekst23: We should definitely get the rest of the team's feedback about how much unit testing we want to integrate. 

You are right in principle, but in theory we should be maintaining our tests together with the app code.

It's a bit of a trade off. Your app code is more rigid, but the application itself will be more resilient to failure.

adr/0005-testing-strategy
alekst23: Ah, spelling. My forteh

adr/0005-testing-strategy
CollectiveUnicorn: The `conftest.py` makes sense to standardize around.

For `unittest.mock.patch` and `pytest.mark.parametrize` I personally would just prefer examples of how to utilize these effectively so they can be easily used where appropriate. As they won't be the best tools for all cases.

alekst23: Yeah that might be getting too granular. I definitely want to set up a template for us to follow through some examples of using these decorators and functions, but it's hard to put that in the ADR.

adr/0005-testing-strategy.md
CollectiveUnicorn: I like the idea of doing it, but I'm worried about its maintanability vs the provided value. There's a lot of overhead involved with trying to enumerate all possible edge cases.

adr/0005-testing-strategy.md
CollectiveUnicorn: Yeah its all a tradeoff, I would prefer that we be diligent in ensuring that we're not just creating tests for the sake of creating test and that they provide clear value in terms of catching regressions, bugs, or serving as documentation.

adr/0005-testing-strategy.md
alekst23: That's true, and I don't want to stretch this concept to absurdity. However, if we don't have at least some coverage of edge cases, we are going to see those come up as runtime errors that may be difficult to find and troubleshoot.

CRUDBase is a good example for this purpose. The code base is very simple, but there are some edge cases, like when the DB fails to return a properly formatted response. If we run those through the tests, we see that the system fails with an abstract TypeError or something like that. However, we may choose to return None or an empty set in some cases, instead of throwing an error, indicating that `get` or `list` did not find anything.

adr/0005-testing-strategy.md
alekst23: Yeah I am not a fan of code coverage just for the sake of code coverage. I do love testing edge conditions though =D

adr/0005-testing-strategy.md
CollectiveUnicorn: That definitely makes sense, given that I think this is good sense I think we're all aware of not going overboard with this.

CollectiveUnicorn (approved):

PR #597, opened 2024-06-06
594 feat(api) assistants helper function to build vector stores

jalling97:
- Adds a helper function to automatically create a vector store when a new assistant is created if files are provided
- Moves create and modify vector store functions into `index.py`
- Adds tests to assistants to catch these new possible scenarios


src/leapfrogai_api/data/async_supabase_vector_store.py
gphorvath, line 1: This file should be fully moved into index.py now.

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -37,16 +46,74 @@ async def create_assistant(
             detail=f"Unsupported tool type: {unsupported_tool.type}",
         )
 
+    # check if the request includes a code interpreter
     if request.tool_resources and any(
-        isinstance(tool_resource, ToolResourcesCodeInterpreter)
-        and tool_resource.get("file_ids")
+        isinstance(tool_resource[1], ToolResourcesCodeInterpreter)
+        and tool_resource[1] is not None
```
YrrepNoj, line 52: This seems pretty complex. Is this effectively the same thing?

```python
if request.tool_resources and request.tool_resources.code_interpreter is not None:
```

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -37,16 +46,74 @@ async def create_assistant(
...
     ):
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
             detail="Code interpreter tool is not supported",
         )
 
+    # check if a vector store needs to be built or added to this assistant
+    if request.tool_resources and any(
+        isinstance(tool_resource[1], ToolResourcesFileSearch)
+        for tool_resource in request.tool_resources
```
YrrepNoj, line 63: 1) Same comment as above about this being a complex-looking statement.

2) Should we be checking that the instance of ToolResourcesFileSearch is not None?

src/leapfrogai_api/routers/openai/assistants.py
jalling97: Yes I believe so, I'll change and test to make sure.

src/leapfrogai_api/routers/openai/assistants.py
jalling97: Added that and verified that works as a check

jalling97:
Proof that all integration tests are passing
Screenshot from 2024-06-13 18-15-54

jalling97:
Refactored a few tests so that I'm no longer using the `pytest.mark.xfail` decorator
Screenshot from 2024-06-14 16-10-04

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -37,16 +42,70 @@ async def create_assistant(
...
+            request.tool_resources.file_search.vector_store_ids = [vector_store.id]
+        elif ids_len == 1:
+            logging.debug(
+                "Attaching vector store with id: {} to new assistant".format(ids[0])
+            )
+            # How do we check to make sure the supplied vector_store_id is valid?
+            crud_vector_store = CRUDVectorStore(db=session)
+
+            try:
+                await crud_vector_store.get(filters=FilterVectorStore(id=ids[0]))
```
YrrepNoj, line 93: Should we be catching the return of this `get()` call and check if it is None? We are raising an exception if an invalid VectorStore ID is provided, but should an expired VectorStore also raise an exception?

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -37,16 +42,70 @@ async def create_assistant(
...
+                    detail="Invalid vector store id was provided",
+                )
+        else:
+            logging.debug(
+                "No files or vector store id found; assistant will be created with no vector store"
+            )
+
+        with suppress(AttributeError):
+            request.tool_resources.file_search.vector_stores = (
+                None  # remove vector_stores if it's there
```
YrrepNoj, line 106: I am probably missing something here, but what is the intuition behind setting this to None?

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -172,6 +227,66 @@ async def modify_assistant(
...
+            request.tool_resources.file_search.vector_store_ids = [vector_store.id]
+        elif ids_len == 1:
+            logging.debug(
+                "Attaching vector store with id: {} to new assistant".format(ids[0])
+            )
+            # How do we check to make sure the supplied vector_store_id is valid?
+            crud_vector_store = CRUDVectorStore(db=session)
+
+            try:
+                await crud_vector_store.get(filters=FilterVectorStore(id=ids[0]))
```
YrrepNoj, line 277: Same comment as above about what if it makes sense to validate against a `None` return here?

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -172,6 +227,66 @@ async def modify_assistant(
...
+            indexing_service = IndexingService(db=session)
+
+            if not indexing_service.file_ids_are_valid(vector_stores[0]["file_ids"]):
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail="Invalid file ids attached to assistants request",
+                )
+
+            vector_store_request = CreateVectorStoreRequest(
+                file_ids=vector_stores[0]["file_ids"],
```
YrrepNoj, line 258: I'm having a hard time following the typing here.

I understand that we expect `len(vector_stores) == 1` so `vector_store[0]` makes enough sense to me. But I'm having a hard time determining what _type_ `vector_store[0]` will return and why we are obtaining a field with `vector_stores[0]["file_ids"]` (and what that _type_ will be as an end result).

src/leapfrogai_api/routers/openai/assistants.py
jalling97: Yeah that's a good catch, I didn't think about expiry situations. I can update the check and wherever else that's relevant

src/leapfrogai_api/routers/openai/assistants.py
jalling97: Based on how Openai does it, it's valid for a createAssistant request to have the `vector_stores` field under `file_search` ([source](https://github.com/openai/openai-python/blob/a33712e161eacdcfead297da3f4a5bc491732e3f/src/openai/types/beta/assistant_create_params.py#L196)), but not after the assistant has been created ([source](https://github.com/openai/openai-python/blob/a33712e161eacdcfead297da3f4a5bc491732e3f/src/openai/types/beta/assistant.py#L22)). 

We're interpreting the reason why to be that once the vector store is created, its `id` gets linked to the assistant, meaning there's no reason to keep the `file_ids` that came with the initial request, as that's not maintained and could get out of sync. So it's best to remove it after the vector store is created, if it exists.

src/leapfrogai_api/routers/openai/assistants.py
jalling97: Updated :+1:

src/leapfrogai_api/routers/openai/assistants.py
jalling97: Updated :+1:

jalling97:
Passing tests from most recent changes
Screenshot from 2024-06-14 17-36-21

src/leapfrogai_api/routers/openai/assistants.py
YrrepNoj: Ok I think that makes sense, a few QoL requests
1) at the start of this elif block, can we set a variable to make it more clear that we are working with a dict, maybe something like `vector_store_dict = vector_stores[0]`
2) It would probably be wise to do some validation on the dict and throw in a `if "file_ids" not in my_dict: raise exception`
   - We have a similar-ish check above, but we would run into an exception before actually calling the `indexing_service.file_ids_are_valid()` function because we'll get a key_error when trying to get the value from the dictionary.
   
   
 3) There is another place above in this file (around line 74) where doing at least `2)` would be useful.

jalling97: Added these :+1:

src/leapfrogai_api/routers/openai/assistants.py
CollectiveUnicorn, line 1: There seems to be a lot of duplication between `create_assistant` and `modify_assistant` are you able to pull out the duplicate code to reduce some of the bloat a bit?

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -37,12 +43,90 @@ async def create_assistant(
...
+    # check if a vector store needs to be built or added to this assistant
+    if request.tool_resources and request.tool_resources.file_search is not None:
+        ids = request.tool_resources.file_search.vector_store_ids
+        vector_stores = (
+            request.tool_resources.file_search.vector_stores
+            if hasattr(request.tool_resources.file_search, "vector_stores")
+            else None
+        )
+
+        ids_len = len(ids) if ids is not None else 0
```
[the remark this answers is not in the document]
jalling97: If `request.tool_resources.file_search` exists, then the `vector_stores_ids` field will also exist based on the pydantic typing. But it's likely that the field is going to be set to `None` (in the event you're providing `file_ids` instead). So that `getattr` would see that the field itself exists but set ids to `None`, which is why that check gets made. There's probably a more efficient way to do it but at some point I'm pretty sure you have to check whether that field is a `None` type or not.

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -37,12 +43,90 @@ async def create_assistant(
...
+        vector_stores_len = len(list(vector_stores)) if vector_stores is not None else 0
```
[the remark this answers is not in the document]
jalling97: This has the same problem that I mentioned with `ids`, but it's still a more efficient way to do it than I was with the `hasattr` check so I agree that this makes more sense. I still believe you have to check to make sure it's not a `None` type so you don't get a `TypeError`.

Additionally, since the OpenAI [assistant_create_params](https://github.com/openai/openai-python/blob/a33712e161eacdcfead297da3f4a5bc491732e3f/src/openai/types/beta/assistant_create_params.py#L205) `ToolResourcesFileSearch` only states that `vector_stores` is an iterable, the list conversion is just a safety check in case an iterator or another `Iterable` gets passed. In the event it is a list, a list conversion on a list is still thankfully a list so it may be slightly more inefficient but it catches those instances.

src/leapfrogai_api/routers/openai/assistants.py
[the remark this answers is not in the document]
CollectiveUnicorn: Yeah, you're right. Maybe `ids = request.tool_resources.file_search.vector_store_ids or []`, but its not a big deal. You can dismiss this if you want.

src/leapfrogai_api/routers/openai/assistants.py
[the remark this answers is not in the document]
CollectiveUnicorn: Probably about the same size wise when accounting for the new try/except block. But I am mostly concerned about readability, so what about swapping this for `0 if vector_stores is None else len(list(vector_stores))` at least so that its affirmative

src/leapfrogai_api/routers/openai/assistants.py
jalling97: I originally pulled out all of the common checks between create and modify and put them as a helper function in `backend/helpers.py`, but that caused a lot of circular import errors so I ended up pulling it out and left it in `assistants.py`. I don't love having a helper function there but aside from making a new utils or helpers file it's currently clearly annotated as a helper function that only applies to `create_assistant` and `modify_assistant`. What do you think?

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -128,6 +136,165 @@ async def index_file(self, vector_store_id: str, file_id: str) -> VectorStoreFil
...
+            if expires_at and expires_at:
+                new_vector_store.expires_after = expires_after
+                new_vector_store.expires_at = expires_at
+
+            return await crud_vector_store.update(
+                id_=vector_store_id,
+                object_=new_vector_store,
+            )
+        except Exception as exc:
+            raise exc
```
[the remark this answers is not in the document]
jalling97: Combined :+1:

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -128,6 +136,165 @@ async def index_file(self, vector_store_id: str, file_id: str) -> VectorStoreFil
...
+                    )
+
+            new_vector_store.status = VectorStoreStatus.COMPLETED.value
+
+            return await crud_vector_store.update(
+                id_=new_vector_store.id,
+                object_=new_vector_store,
+            )
+        except Exception as exc:
+            raise exc
```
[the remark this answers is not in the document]
jalling97: Combined :+1:

src/leapfrogai_api/routers/openai/assistants.py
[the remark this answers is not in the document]
jalling97: I ended up combining the two approaches:
```
ids = request.tool_resources.file_search.vector_store_ids or []
vector_stores = (
    getattr(request.tool_resources.file_search, "vector_stores", []) or []
)
```
Since `vector_store_ids` will always exist but `vector_stores` may not.

src/leapfrogai_api/routers/openai/assistants.py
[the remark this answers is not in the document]
jalling97: As stated above, I ended up combining the two approaches to testing for `None` types so now I just do a basic length check
```
ids_len = len(ids)
vector_stores_len = len(list(vector_stores))
```

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -217,6 +380,20 @@ async def asimilarity_search(self, query: str, vector_store_id: str, k: int = 4)
...
+    async def _increment_vector_store_file_status(
+        self, vector_store: VectorStore, file_response: VectorStoreFile
+    ):
+        """Increment the file count of a given vector store based on the file response"""
+        if file_response.status == VectorStoreFileStatus.COMPLETED.value:
+            vector_store.file_counts.completed += 1
+        elif file_response.status == VectorStoreFileStatus.FAILED.value:
+            vector_store.file_counts.failed += 1
+        elif file_response.status == VectorStoreFileStatus.IN_PROGRESS.value:
+            vector_store.file_counts.in_progress += 1
```
YrrepNoj, line 392: Does the `in_progress` counter ever get decremented? It seems like this will be a forever growing count.

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -224,3 +201,103 @@ async def delete_assistant(
...
+    if request.tools:
+        for tool in request.tools:
+            if not validate_assistant_tool(tool):
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail=f"Unsupported tool type: {tool.type}",
+                )
+
+    # check for unsupported tool resources
+    if not validate_tool_resources(request.tool_resources):
```
YrrepNoj, line 222:
```suggestion
    if request.tool_resources and not validate_tool_resources(request.tool_resources):
```

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -224,3 +201,103 @@ async def delete_assistant(
         deleted=bool(assistant_deleted),
         object="assistant.deleted",
     )
+
+
+# helper function to check for unsupported tools/tool resources, and if a new vector store needs to be added
+async def _assistant_request_checks(
```
YrrepNoj, line 207: I like the spirit of this helper function and think it is in the right direction. I would suggest splitting the logic in this function up more to make it easier to read. It might be worthwhile to put the sections under the `elif vector_store_len == 1` and `elif ids_len == 1` into their own sub-functions.

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -224,3 +201,103 @@ async def delete_assistant(
...
+    request: CreateAssistantRequest | ModifyAssistantRequest, session: Session
+) -> CreateAssistantRequest | ModifyAssistantRequest:
+    """Performs common checks that occur in both create and modify requests"""
```
YrrepNoj, line 210: I love the addition of the doc string. This is a pattern we should be doing more often.

I want to challenge this docstring a little though. After reading this docstring (and function name), I expect this function to ONLY perform checks and return a bool/status. I wouldn't expect this function to be modifying the input and returning it back to the caller.

We should make the docstring (and maybe even the function name) a little more clear that this is happening.

src/leapfrogai_api/routers/openai/assistants.py
```diff
@@ -224,3 +201,103 @@ async def delete_assistant(
...
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail="Invalid vector store id was provided",
+                )
+        else:
+            logging.debug(
+                "No files or vector store id found; assistant will be created with no vector store"
+            )
+
+        with suppress(AttributeError):
+            request.tool_resources.file_search.vector_stores = (
```
YrrepNoj, line 299: I am assuming this is purposeful because of the `with suppress()` in the line above, but are we sure we don't want to be modifying `file_search.vector_store_ids`?

YrrepNoj (changes requested):

src/leapfrogai_api/routers/openai/assistants.py
jalling97: Good question. So in the event that the request has a `file_search.vector_stores` field, we create a new vector store and link that id to the request instead. This happens on [line 273](https://github.com/defenseunicorns/leapfrogai/blob/b42646f14484d59b330b5657df0d934758543cfd/src/leapfrogai_api/routers/openai/assistants.py#L273) at the moment. So we either start with `vector_stores` and end with `vector_store_ids` or start with `vector_store_ids` and end with the same. The only reason this line to set `vector_stores` to None is outside everything is to ensure nothing sneaks through.

Although I could probably handle that by removing the `vector_stores` within the same portion of code as creating the vector store and just making an `assertion` at the end. Whichever is preferred.

src/leapfrogai_api/backend/rag/index.py
jalling97: After checking with @gphorvath , no, it currently has the initial pass then remains unchanged for the life of the vector store, which is a problem. Would it make sense to open another issue for this bug? Or do you just want me to include a solution in this PR?

src/leapfrogai_api/routers/openai/assistants.py
jalling97: Changed :+1:

src/leapfrogai_api/routers/openai/assistants.py
jalling97: Updated :+1:

src/leapfrogai_api/backend/rag/index.py
jalling97: This will be addressed in the following issue: https://github.com/defenseunicorns/leapfrogai/issues/743

YrrepNoj (approved):
LGTM. Thanks for the consistent updates.

CollectiveUnicorn (approved):

YrrepNoj (approved):

PR #606, opened 2024-06-11
fix(api): Migrations

gphorvath:
* Moves vector stores migrations into `packages/api/migrations`
* Moves `supabase` into `src/leapfrogai_api/supabase` since it's only for the API
* Adds a symbolic link between `packages/api/migrations` and `src/leapfrogai_api/supabase/migrations`


YrrepNoj (approved):
I don't love the idea of introducing a symlink, but I understand the motivation and reasoning behind doing it.

I'll approve this for now but I will be brainstorming other ways we can do this for the near future.

jalling97 (approved):

PR #629, opened 2024-06-14
573 api integration

### LINKED ISSUE #573: chore(ui): Integrate UI with new LF API
Swap the env vars for the frontend to connect to the new LF API. 

Ensure auth is integrated with the user's session token.

andrewrisse:
closes #573 

Integrates with new LF API and still works with OpenAI.


src/leapfrogai_ui/src/lib/components/AssistantForm.svelte
```diff
@@ -78,7 +78,7 @@
   });
 
   $: if (data.files) {
-    const fileRows: FileRow[] = data.files.map((file) => ({
+    const fileRows: FileRow[] = data.files?.map((file) => ({
```
YrrepNoj, line 81: Ignorant question;

My understanding of the `?` operator is that it does a 'null/undefined' check against `data.files` before running the mapping function. Is that correct? And if that is correct, isn't that handled by the above `if (data.files)` statement?

src/leapfrogai_ui/src/routes/api/chat/+server.ts
```diff
@@ -28,11 +23,10 @@ export async function POST({ request, locals: { safeGetSession } }) {
   } catch {
     error(400, 'Bad Request');
   }
-
-  // Add the default system prompt to the beginning of the messages
-  if (messages[0].content !== env.DEFAULT_SYSTEM_PROMPT) {
-    messages.unshift({ content: env.DEFAULT_SYSTEM_PROMPT!, role: 'system' });
-  }
+  const openai = createOpenAI({
```
YrrepNoj, line 26: I've noticed other sections of this PR use `src/lib/server/constants.ts -> getOpenAiClient()`. Does it make sense to use that helper here instead?

YrrepNoj (dismissed):

src/leapfrogai_ui/src/lib/components/AssistantForm.svelte
andrewrisse: it sure does, removing

src/leapfrogai_ui/src/routes/api/chat/+server.ts
andrewrisse: this is a special case where the vercel ai sdk expects a different version of the openai client so we create it with their helper (for streaming). I will add a comment

YrrepNoj (approved):

PR #630, opened 2024-06-14
chore: Handle migrations with least permissions necessary

YrrepNoj:
This PR:
1) Consolidates the Dockerfiles that we use when running migrations
2) Creates a new component within the supabase package that runs migrations that require elevated (`supabase_admin`) permissions so that the other migration jobs can run as the regular `postgres` user.


Makefile
```diff
@@ -68,7 +73,7 @@ build-ui: local-registry ## Build the leapfrogai_ui container and Zarf package
 	docker tag ghcr.io/defenseunicorns/leapfrogai/leapfrogai-ui:${LOCAL_VERSION} localhost:5000/defenseunicorns/leapfrogai/leapfrogai-ui:${LOCAL_VERSION}
 
 	## Build the migration container for the version of the UI
-	docker build -t ghcr.io/defenseunicorns/leapfrogai/ui-migrations:${LOCAL_VERSION} src/leapfrogai_ui/supabase
+	docker build -t ghcr.io/defenseunicorns/leapfrogai/ui-migrations:${LOCAL_VERSION} -f Dockerfile.migrations --build-arg="MIGRATIONS_DIR=src/leapfrogai_ui/supabase/migrations" .
```
YrrepNoj, line 76: What do we think about the tag for these images? I was thinking about if I like having the images explicitly listed out for the different packages or if I wanted to have a more common name. where the ref will specify which package it is for.

IE)
`ghcr.io/defenseunicorns/leapfrogai/lfai-supabase-migrations:0.8.0`
`ghcr.io/defenseunicorns/leapfrogai/api-migrations:0.8.0` 
`ghcr.io/defenseunicorns/leapfrogai/ui-migrations:0.8.0` 

vs
`ghcr.io/defenseunicorns/leapfrogai/lfai-migrations:0.8.0-base`
`ghcr.io/defenseunicorns/leapfrogai/lfai-migrations:0.8.0-ui`
`ghcr.io/defenseunicorns/leapfrogai/lfai-migrations:0.8.0-api`

justinthelaw: The advantage of the former is that release-please will play more easily when tagged versions are semver compliant. The 2nd one is nice since it feels more flavor-like.

My vote is on the second one, despite some release-please shenanigans that will have to happen.

justinthelaw (dismissed):

gphorvath (dismissed):

jalling97 (dismissed):

jalling97 (approved):

PR #632, opened 2024-06-14
(feat) Tests for Threads and Runs

alekst23:
Mocks for db and auth
Parameterization of test calls


jalling97:
I see the lint check isn't passing right now; are you using the pre-commit hooks when doing local dev by any chance?

jalling97:
the `pyproject.toml` at the [root of the repo](
image
) is probably a good place for the missing `dotenv` dependency to live, since that's where we've been putting base dev/ci dependencies.

PR #677, opened 2024-06-20
(tests) Unit Testing for CRUDBase

alekst23:
Creates mocks for db session
Creates unit testing for CRUDBase features

Current Errors:
Most sad paths fail because we do not properly process empty return types from the database

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,8 @@ dependencies = [  # Dev dependencies needed for all of lfai
     "pytest",
     "httpx",
     "ruff",
+    "python-dotenv",
+    "pytest-asyncio"
 ]
 requires-python = "~=3.11"
 
--- a/src/leapfrogai_api/data/crud_base.py
+++ b/src/leapfrogai_api/data/crud_base.py
@@ -29,7 +29,7 @@ class CRUDBase(Generic[ModelType]):
             del dict_["created_at"]
 
         result = await self.db.table(self.table_name).insert(dict_).execute()
-        
+
         try:
             return self.model(**result.data[0])
         except Exception:
--- a/src/leapfrogai_api/data/crud_base.py
+++ b/src/leapfrogai_api/data/crud_base.py
@@ -23,15 +23,17 @@ class CRUDBase(Generic[ModelType]):
         ):  # There are cases where the id is provided
             del dict_["id"]
         # Only delete created_at if it is <= 0, the db time is not adequate for message ordering
-        if "created_at" in dict_ and dict_["created_at"] <= 0:
+        if "created_at" in dict_ and not (
+            isinstance(dict_["created_at"], int) and dict_["created_at"] > 0
+        ):
             del dict_["created_at"]
-        data, _count = await self.db.table(self.table_name).insert(dict_).execute()
 
-        _, response = data
-
-        if response:
-            return self.model(**response[0])
-        return None
+        result = await self.db.table(self.table_name).insert(dict_).execute()
+        
+        try:
+            return self.model(**result.data[0])
+        except Exception:
+            return None
 
     async def get(self, filters: dict | None = None) -> ModelType | None:
         """Get row by filters."""
@@ -41,13 +43,12 @@ class CRUDBase(Generic[ModelType]):
             for key, value in filters.items():
                 query = query.eq(key, value)
 
-        data, _count = await query.execute()
+        result = await query.execute()
 
-        _, response = data
-
-        if response:
-            return self.model(**response[0])
-        return None
+        try:
+            return self.model(**result.data[0])
+        except Exception:
+            return None
 
     async def list(self, filters: dict | None = None) -> list[ModelType] | None:
         """List all rows."""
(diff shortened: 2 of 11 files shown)


make-tests.mk
```diff
@@ -0,0 +1,6 @@
+
+test-api-integration:
```
jalling97, line 2: What's the benefit of adding another makefile as opposed to putting these make targets in the root Makefile?

tests/conftest.py
```diff
@@ -0,0 +1,13 @@
+import os
+from dotenv import load_dotenv
+
+
```
jalling97, line 4: Don't pytest fixtures normally live in `conftest.py` as well?

tests/conftest.py
gphorvath, line 1: I believe this conftest.py is causing the Supabase test to fail in the e2e pipeline. If possible we should probably try to limit the scope of mocks to the unit test directory to avoid mocking something that is used in an e2e or integration test.

https://github.com/defenseunicorns/leapfrogai/actions/runs/9617886637/job/26530618578#step:9:113

make-tests.mk
gphorvath: I'm personally a fan of starting to break out the Makefile in this way to avoid a massive Makefile. Although it might be nice if we can move the includes into a directory to avoid cluttering up the root directory.

pyproject.toml
gphorvath, line 1: I'm also seeing pytest-asyncio as a requirement for the unit tests.

make-tests.mk
```diff
@@ -0,0 +1,6 @@
...
+	pytest tests/integration/api
```
gphorvath, line 3: More a point of preference, but I've been starting to add `-vv -s` to get more verbosity and detail when things fail.

src/leapfrogai_api/data/crud_base.py
```diff
@@ -23,15 +23,17 @@ async def create(self, object_: ModelType) -> ModelType | None:
...
-        _, response = data
+        result = await self.db.table(self.table_name).insert(dict_).execute()
 
-        if response:
-            return self.model(**response[0])
-        return None
+        try:
+            return self.model(**result[0][1][0])
+        except Exception:
+            return None
```
gphorvath, line 36: This seems to be causing errors. It looks like we can't access the `data` element by using `[0]` index. I found this works, but maybe you can find a better way? This is similarly an issue in list and delete endpoints. Might be worth running the integration tests periodically to make sure nothing is breaking.

```
        result, _ = await self.db.table(self.table_name).insert(dict_).execute()

        try:
            return self.model(**result[1][0])
        except Exception:
            return None
```

gphorvath (changes requested):

make-tests.mk
alekst23: @jalling97 The benefit is that we have better organized code instead of a monolithic Makefile.

@gphorvath I like the idea of organizing them by directory. There are two options to do this:
1) One directory for all Makefile includes, like `common` in project root.
2) Keep Makefiles closer to their target, in the relevant component directory.

I think option (2) may be better for Makefiles specific to a module or function, like Testing, Supabase, etc.

pyproject.toml
alekst23: Added to `project.toml`

src/leapfrogai_api/data/crud_base.py
alekst23: That's strange. Is there error somewhere up stream in integration testing or elsewhere?

My unit tests are running OK for this. Perhaps I am not mocking the db response properly.

tests/conftest.py
alekst23: You can have multiple conftest.py files, and they will be included for all tests in that directory and lower. These files usually have the relevant fixtures defined therein.

So this is the root conftest for all tests. The fixtures we define elsewhere in this PR are specifically for unit testing, and I did not want to them to be loaded for all tests like integration and e2e.

tests/conftest.py
alekst23: This is an interesting problem. Our conftest file should be using environment variables where they have been defined, and using mock data where they are not.

I think the problem may be that we don't have ANON_KEY in env? It seems to be resolving this url https://supabase-kong.uds.dev/auth/v1/health but fails with 401

make-tests.mk
alekst23: I went with option (2) and moved the file to `tests/make-tests.mk`

Let me know if you feel better about a common directory for all makefiles, but this seems to make sense to me for now.

src/leapfrogai_api/data/crud_base.py
alekst23: @gphorvath I think I found the problem. It's related to `table.select` returning a list, wheres all other methods return objects. I have made the relevant fixes to mocks and unit tests, but I am having some trouble running the integration tests right now to verify.

gphorvath (approved):
Looks good to me now! Can you make sure we have an issue tracking adding these tests to a pipeline?

PR #699, opened 2024-07-01
fix(api): bump max tokens to maximum for Synthia-7B to prevent chopping

gphorvath:
This should fix the issue with chopping in the UI when using the default model `TheBloke/Synthia-7B-v2.0-GPTQ`. Bumps the default max tokens to 4096.

### DIFF
--- a/src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py
+++ b/src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py
@@ -72,7 +72,7 @@ class RunCreateParamsRequestBase(BaseModel):
     assistant_id: str = Field(default="", examples=["123ab"])
     instructions: str = Field(default="", examples=["You are a helpful AI assistant."])
     max_completion_tokens: int | None = Field(default=4096, examples=[4096])
-    max_prompt_tokens: int | None = Field(default=4096, examples=[32768])
+    max_prompt_tokens: int | None = Field(default=4096, examples=[4096])
     metadata: dict | None = Field(default={}, examples=[{}])
     model: str | None = Field(default=None, examples=["llama-cpp-python"])
     response_format: AssistantResponseFormatOption | None = Field(
--- a/src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py
+++ b/src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py
@@ -67,12 +67,19 @@ from leapfrogai_sdk.chat.chat_pb2 import (
     ChatCompletionResponse as ProtobufChatCompletionResponse,
 )
 
+DEFAULT_MAX_COMPLETION_TOKENS = 4096
+DEFAULT_MAX_PROMPT_TOKENS = 4096
+
 
 class RunCreateParamsRequestBase(BaseModel):
     assistant_id: str = Field(default="", examples=["123ab"])
     instructions: str = Field(default="", examples=["You are a helpful AI assistant."])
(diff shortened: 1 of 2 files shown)


justinthelaw (approved):

src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py
```diff
@@ -98,12 +98,12 @@ def __init__(self, **data):
             logging.warning(
                 "max_completion_tokens is not set or is less than 1, setting to 1024"
             )
-            self.max_completion_tokens = 1024
+            self.max_completion_tokens = 4096
         if self.max_prompt_tokens is None or self.max_prompt_tokens < 1:
             logging.warning(
                 "max_prompt_tokens is not set or is less than 1, setting to 1024"
```
YrrepNoj, line 104: The log statement is now incorrect. Maybe we should make a CONST for the default value so that we can use the const in this logged string as well?

YrrepNoj (changes requested):
One small callout about a bad log statement.

src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py
gphorvath: Good catch, fixed

src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py
```diff
@@ -67,12 +67,19 @@
     ChatCompletionResponse as ProtobufChatCompletionResponse,
 )
 
+DEFAULT_MAX_COMPLETION_TOKENS = 4096
```
justinthelaw, line 70: Now that @YrrepNoj pointed it out, this might be best exposed as a Zarf variable. I assume this really only affects the UI, where we cannot modify or supply this variable via API. If a delivery engineer chooses to use a larger model (like at BeastCode, with Mixtral) with larger context or a specific use-case that requires more generation, then they would want to be able to change this.

justinthelaw: At the same time, you can feel free to say that this goes beyond the purpose of the ticket.

gphorvath: In theory this is only the default value and if a different value is passed in via the API it will be overridden.

gphorvath: I think for now we can just allow it to be overridden via API and in the future we handle at the time we load a model in via the model registry. We should be creating custom configurations that are loaded with each new model.

justinthelaw (approved):

CollectiveUnicorn (approved):

src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py
```diff
@@ -67,12 +67,19 @@
...
+DEFAULT_MAX_PROMPT_TOKENS = 4096
```
YrrepNoj, line 71: nit; what do you think about moving this to the types file so we can reuse this in the `CompletionRequest.max_tokens` field?

YrrepNoj (approved):
One small nit; approving and leaving it up to you if you want to resolve or not. 🫶

src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py
gphorvath: I took your suggestion!

YrrepNoj (approved):

PR #713, opened 2024-07-02
chore(ui) - fix playwright e2e tests

andrewrisse:
After transition to Leapfrog API, our E2Es started having more failures. This PR improves the stability of the tests and fixes several bugs.

- Fix file upload success indicator hiding

- Fix disable checkbox for file uploading

- Improve sorting logic and fix disable send (rate limiting logic)

- Create unique pdfs for file-management tests 

- Refactor to handle LF API not allowing file read stream yet (for file uploads)

- Re-enable skipped tests that were flaky

- Add cleanup step at end of tests (reduces flakiness when test files were trying to clean up when other files were still running)

- Refactor to 1 worker and only chromium - running more than 1 worker can cause flakiness due to test files being run at the same time in different browsers (e.x. navigation history is incorrect). Additionally, Leapfrog API is slow when attaching files to assistants, resulting in flaky tests. We can attempt in increase number of browser and workers in the pipeline when the API is faster.


src/leapfrogai_ui/playwright.config.ts
```diff
@@ -1,44 +1,57 @@
...
+//     storageState: 'playwright/.auth/user.json'
+//   },
+//   dependencies: ['setup']
+// };
+// const webkitConfig = {
+//   name: 'webkit',
+//   use: { ...devices['Desktop Safari'], storageState: 'playwright/.auth/user.json' },
+//   dependencies: ['setup']
+// };
+// const edgeConfig = {
```
justinthelaw, line 32: Is there a benefit to running tests on Edge, or are the browser environments so similar (relative to Chrome) that it isn't required? Base don the docs, it seems like the Edge tests would run concurrently, and not use much compute overhead in the Playwright environment. Same comment for Safari.

The reason I ask is because most of our future users will probably use the UI in Chrome(ium), Safari, and/or Edge, since they are government or developers.

src/leapfrogai_ui/src/lib/mocks/supabase-mocks.ts
```diff
@@ -28,7 +28,7 @@ export const sessionMock = vi.fn(() => {
       email,
       email_verified: true,
       full_name,
-      iss: 'https://keycloak.admin.uds.dev/realms/uds',
```
justinthelaw, line 31: Just for the sake of completeness, can and should we load and input the ENV for this auth callback URL instead?

src/leapfrogai_ui/playwright.config.ts
andrewrisse: There is definitely benefit, but I have had two main issues when running multiple browsers:
1. The API (especially when attaching files to assistants) slows down too much when multiple browsers are running multiple tests against it. This causes lots of test flakiness and failures. I'm hoping when we handle concurrent requests better, we can re-enable more browsers.
2. A couple tests can cause issues if two browsers run them at the same time. For example, I noticed one of our tests specifically checks navigation history, and multiple browsers running this test at the same time causes flakiness. I didn't think the different browsers would affect the navigation history, but I saw this issue consistently and ran out of time to debug.

Because of these, I've elected to run the tests in one browser for now so we have passing tests and can finish the pipeline integration story. These days, it's a lot more rare that browsers behave differently (although it does happen!), so I feel comfortable only running one browser for now until we have time for test performance improvements.

src/leapfrogai_ui/src/lib/mocks/supabase-mocks.ts
andrewrisse: updated!

src/leapfrogai_ui/playwright.config.ts
justinthelaw: That is super reasonable to me. Are there issues already written for the problems you've encountered with the flakiness caused by API slow-down? There should be a way to more gracefully handle the slowness, either by slowing down the test and awaiting for an event (I only remember how to do this in Cypress...); or this could be purely API-side, although the difficulty is that our runner will always experience fluctuation in compute time due to it being a hosted resource.

andrewrisse: I will make an issue to re-address , but yes I am also concerned about the runners. They tend to be very slow on e2e tests. We can test further  when we complete this story (I'll add a comment): https://github.com/defenseunicorns/leapfrogai/pull/551

You can adjust timeouts, and I've messed with that, but they were getting excessive. Running the tests for each browser concurrently (disable parallelism), did work, but they were extremely slow. This may be less of a problem when we run them headless, but the pipeline may run into issues.

andrewrisse: https://github.com/defenseunicorns/leapfrogai/issues/742

justinthelaw (approved):

PR #748, opened 2024-07-09
chore(api): raise exception on CRUD base fails

alekst23:
CRUD operations will now raise exceptions instead of catching them and returning None.
- Will return None for exceptions that are due to no results from db

Delete function will only return True/False.

### DIFF
--- a/src/leapfrogai_api/data/crud_base.py
+++ b/src/leapfrogai_api/data/crud_base.py
@@ -78,6 +78,7 @@ class CRUDBase(Generic[ModelType]):
         except Exception as exc:
             raise Exception from exc
 
+
     async def update(self, id_: str, object_: ModelType) -> ModelType | None:
         """Update a row by its ID."""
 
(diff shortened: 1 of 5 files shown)


src/leapfrogai_api/data/crud_base.py
gphorvath, line 1: @alekst23 now that we are reraising exceptions, we should make sure to catch the index out of range error inside of crud_base and return none or false respectively.

image

gphorvath (changes requested):

YrrepNoj (approved):

src/leapfrogai_api/data/crud_base.py
alekst23: Good idea. This is now implemented.

YrrepNoj (approved):
LGTM (after conflict is resolved).

gphorvath (approved):

Issue #746, opened 2024-07-09
spike: Investigate runtime and file queueing


CollectiveUnicorn:
### Describe what should be investigated or refactored
The following issues are aimed at creating queueing solutions to deal with high amounts of traffic: https://github.com/defenseunicorns/leapfrogai/issues/482 https://github.com/defenseunicorns/leapfrogai/issues/626 when performing various tasks. Investigate potential solutions and architectural changes. Document your findings in an ADR.

PR #749, opened 2024-07-09
spike: Investigate runtime and file queueing

CollectiveUnicorn:
- Adds an ADR to determine how we should go about handling performance bottlenecks, in addition to what tools we should use if any for queuing.
- Closes: https://github.com/defenseunicorns/leapfrogai/issues/746


adr/0006-queueing-high-traffic.md
```diff
@@ -0,0 +1,40 @@
...
+  - [References](#references)
+
+## Status
+
+PROPOSED
+
+## Context
+
+LeapfrogAI has to be able to handle a large volume of inference, file upload, and embeddings requests. To ensure that we can handle this level of activity without significant performance degradation we need to have systems in place to prevent the system from being overwhelmed or blocked by a large volume or a single long running task. 
+
```
alekst23, line 22:
```suggestion
Adding a Queue management component can help us create a more efficient request management system to deal with high request volumes of long-running tasks. However, it introduces a significant level of complexity to the system, and we must weight the options carefully.

### Benefits of Queue system for request processing:
- Allows the Gateway to respond even when the back-end is very busy
- Prevents request from being dropped or timing out
- Allows resuming failed requests
- Allows to throttle message processing rate
```

CollectiveUnicorn: This is good, though needs to be weigh instead of weight

barronstone:
The rationale makes sense to me at a high level: Take a lightweight approach for now and take the more complex path later if necessary.

gphorvath (approved):

gphorvath:
This makes sense as `Proposed`. We need to make sure to revisit as we begin implementing.

gphorvath (approved):

PR #759, opened 2024-07-10
Updating default GPU resources

unicorn-kp:
Updated uds-config.yaml to default to a single GPU to ease the use of quick starts. This ensures that users with a single GPU can deploy out of the box.

Updated the docs to include a disclaimer about this and instructions to add additional GPU resources.


YrrepNoj (approved):
This seems like a reasonable enough change. The UDS bundles that we define here are merely examples so it makes sense to build the examples to the more common use case of people only having a single GPU.

LGTM

justinthelaw (approved):

PR #768, opened 2024-07-12
feat(api): set api variable to expose by default

gphorvath:


uds-bundles/dev/README.md
```diff
@@ -11,7 +11,11 @@ python -m venv .venv
 source .venv/bin/activate
 ```
 
-Each component is built into its own Zarf package. You can build all of the packages you need at once with the following `Make` targets:
+Each component is built into its own Zarf package.
+
+| ***Note:*** You need to build with `make build-* LOCAL_VERSION=dev` to set the tag to `dev` instead of the commit hash locally.
```
YrrepNoj, line 16: Was this supposed to be a quote block? 

```suggestion
> ***Note:*** You need to build with `make build-* LOCAL_VERSION=dev` to set the tag to `dev` instead of the commit hash locally.  
```

gphorvath: why yes it is

YrrepNoj (dismissed):
LGTM

YrrepNoj (approved):

PR #773, opened 2024-07-12
fix(api): type-checking embedding list[str]

gphorvath:


CollectiveUnicorn (approved):

src/leapfrogai_api/routers/openai/embeddings.py
```diff
@@ -29,12 +29,25 @@ async def embeddings(
...
+    new_list: list[str] = []
+    for item in v:
+        if not isinstance(item, str):
+            raise HTTPException(
+                status_code=status.HTTP_405_METHOD_NOT_ALLOWED,
+                detail=f"Invalid input type {type(item)}. Currently supported types are str and list[str]",
+            )
+
+        new_list.append(item)
+    return new_list
```
jalling97, line 53: Can this helper function get pulled into the request type? Similarly to how the request types 
 [here](https://github.com/defenseunicorns/leapfrogai/tree/main/src/leapfrogai_api/routers/openai/requests) were updated so that the router file doesn't have non-endpoint functions in it?
 
 If not then I do think adding the helper function here detracts more than getting rid of the red squiggly adds, but happy to discuss that.

gphorvath: I think refactoring our request validation into a separate file makes sense, but feels a bit more scope than the intent of this bug fix. I'm personally on team eliminate red squiggles though, because that isn't less readable in my opinion.

jalling97: I could be wrong, but I think we can resolve the red squiggly without the helper function by just casting it to a list, right? i.e:
```
request = lfai.EmbeddingRequest(inputs=list(req.input))
```
If `req.input` is an iterable, it ensures it becomes a list. If it's a list, then nothing happens. Again, could be wrong on that or missing something from the problem setup.

gphorvath: the issue is that list[int] is incompatible with list[str]. list and iterable are compatible

jalling97: Ah I see. So long as we go back and move this help function around later then I'm cool with this

jalling97 (approved):

PR #775, opened 2024-07-12
chore(supabase): Increases supabase-storage file size and makes it configurable

CollectiveUnicorn:
- Increases the file upload size to 512Mb from 50Mb in supabase-storage. This size limits how large the files are that we can use for file uploads and by proxy RAG.
- Adds Zarf variable to configure the value at deploy time.
- 512Mb matches what OpenAI allows for the files upload endpoint


packages/supabase/zarf.yaml
```diff
@@ -49,6 +49,9 @@ variables:
   - name: ENABLE_EXTERNAL_KEYCLOAK
     description: 'Enable the integration between Supabase and an external keycloak instance be enabled'
     default: "true"
+  - name: MAX_FILE_UPLOAD_SIZE
```
YrrepNoj, line 52: Thoughts on putting the units in the name?

packages/supabase/zarf.yaml
```diff
@@ -49,6 +49,9 @@ variables:
...
+    description: 'The maximum upload file size for Supabase storage'
```
justinthelaw, line 53: Maybe a unit would be good here `in bytes, default is equivalent to 100Mb`

packages/supabase/bitnami-values.yaml
```diff
@@ -101,6 +101,9 @@ storage:
   resourcesPreset: "none"
   podLabels:
     sidecar.istio.io/inject: "false"
+  extraEnvVars:
+    - name: FILE_SIZE_LIMIT
```
YrrepNoj, line 105: Maybe a silly question but do you have a link to where this ENV_VAR is loaded? I saw the supabase docs for setting this in the `config.toml` but my quick google search on my phone didn't show anything about env vars.

CollectiveUnicorn: Here's the documentation for it: https://supabase.com/docs/guides/self-hosting/storage/config

I can't say 100% this is where they do it, but it seems to get loaded here: https://github.com/supabase/storage/blob/f6d2fbcdf12d60d9fc8659c0a099c6714431f358/src/config.ts#L211

packages/supabase/zarf.yaml
CollectiveUnicorn: Added this message to the description

packages/supabase/bitnami-values.yaml
YrrepNoj: From that second link you provided; What do you think about using the primary ENV_VAR name `UPLOAD_FILE_SIZE_LIMIT` instead of the fallback name of `FILE_SIZE_LIMIT`?

gphorvath (approved):

packages/supabase/bitnami-values.yaml
CollectiveUnicorn: We can do that, I will say though that given that their public facing documentation says `FILE_SIZE_LIMIT` I would expect that to be less likely to change than `UPLOAD_FILE_SIZE_LIMIT`. But we could probably just set them both.

YrrepNoj (dismissed):
LGTM

packages/supabase/bitnami-values.yaml
```diff
@@ -87,6 +87,11 @@ storage:
...
```
CollectiveUnicorn, line 91: Add comment on why this necessary

gphorvath (approved):

PR #786, opened 2024-07-16
chore: Remove model weights from container images

YrrepNoj:
This PR is a proof of concept to how we can utilize [Zarf Data Injections](https://docs.zarf.dev/ref/components/#data-injections) to push model weights to a [PVC](https://kubernetes.io/docs/concepts/storage/persistent-volumes/) instead of directly building them into the Docker images themselves.

Pros:
 - Much smaller image size (but similar Zarf Package sizes; with the exception of whisper which is now smaller as well)

| Model Container | Original Size | New Size |
| ---- | ---- | ---- |
| llama-cpp-python | 4.58 GB | 0.21 GB |
| whisper | 5.98 GB | 0.63 GB |
| text-embeddings | 10.4 GB | 5.63 GB |
| vLLM | 14.2 GB | 9.62 |

 - Faster build & deploy time
    - I did not quantify this bullet by testing the timings locally but I noticed significant speedups as I was performing other tests locally. 
    - As a data-point, the e2e tests from this branch are running the model creation & deployment steps an average of 2 minutes faster than the tests branching from `main`. This is about a 30% speedup for this workflow step.

<br><br>

Cons:
- Adds a tiny bit of complexity to our deployment by introducing a new PVC. This seems pretty reasonable, but it is still worth noting that this is an added thing that delivery engineers would have to manage (for the storage class of the PVC).
- Adds a dependency on Python (and the `dev` optional-dependencies listed in the `pyproject.toml` being installed on the host machine during the `zarf package create` lifecycle.
    - This dependency is now introduced because we are now pulling down the model weights during the Zarf package creation time instead of pulling the model weights during the Docker container build time.


packages/llama-cpp-python/zarf.yaml
```diff
@@ -26,3 +26,22 @@ components:
...
+          namespace: leapfrogai
+          selector: app=lfai-llama
+          container: data-loader
+          path: /data/.model
+        compress: true
+    actions:
+      onCreate:
+        before:
+          # NOTE: This assumes python is installed and in $PATH
+          - cmd: python scripts/model_download.py
```
YrrepNoj, line 42: I am not completely in love with using a python script here. 

I like the concept of the Zarf package regulating the existence of the model weights (verses the package just assuming the model has already been downloaded and exists in a certain location) but I don't know if a Python script is the best way to do this. This adds a dependency of python at the package creation time.


Any thoughts or spears? Anyone particularly like this approach?

YrrepNoj: I kept it as python for now just to reuse what we currently had for the model_download in the Docker build step, but maybe just making this a bash script makes more sense in the long term.

justinthelaw: The python script was an artifact of wanting to use `hf_transfer`, the rust plugin for maxing out download bandwidth. `hf_transfer` usually only works for the `hugging_face[cli]` version before latest, so that may be why the model download speed differences aren't apparent. Alos, Synthia-7b is a small sized model, relative to the Mixtral variants we were having trouble downloading in a timely manner (before the HF mirror serving the model download would hang on us).

I don't mind reverting this back to a `wget` or `curl` though, if we want to minimize dependencies to only UDS CLI..

packages/llama-cpp-python/chart/templates/pv.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+    name: lfai-llama-pv-claim
```
CollectiveUnicorn, line 4: Are we going to have separate PVs for llama-cpp and vllm?

CollectiveUnicorn (changes requested):

packages/llama-cpp-python/chart/templates/pv.yaml
YrrepNoj: I think yes

CollectiveUnicorn (approved):

justinthelaw:
The Dockerfiles still contain a lot of extra ENVs and ARGs. I recommend removal and an update of the README to reflect the changes made here.

packages/vllm/vllm-values.yaml
```diff
@@ -3,3 +3,8 @@ image:
 
 gpu:
   runtimeClassName: nvidia
+
+persistence:
+  size: 15Gi
+  accessModes: ReadWriteOnce
+  storageClass: "local-path"
```
YrrepNoj, line 10: @justinthelaw Right now these values are not exposed as a Zarf variable. I set the defaults to what I think should be reasonable for a single node cluster.

From your experience, do you think these would make sense to expose as a Zarf Variable or do you think having future deployers rely on UDS Overrides is good for this configuration?

justinthelaw: There are situations where we have had to create 1 or more custom backends, since we usually do not try to deploy the default model that we publish to GHCR. Exposing these at a Zarf package level allows us more flexibility, while still letting us access create or deploy-time values via UDS overrides.

My opinion is to expose these as Zarf variables to accommodate changes at a package level, as we may need 1 or more changes and models in a given cluster.

justinthelaw (approved):

PR #787, opened 2024-07-16
feat(backend): nvidia runtimeclass

gphorvath:
Updates the text-embeddings, whisper, and vllm packages and UDS bundles based on the new runtimeClass for nvidia containers. This goes along with the following PR to update our k3d-gpu base image / daemonset: https://github.com/defenseunicorns/leapfrogai/pull/797


justinthelaw (dismissed):
Tested new `spec.runtimeClassName: nvidia` value on live RKE2 cluster with NVIDIA GPU Operator, which uses same k8s-device-plugin in the custom K3d K3s image. Everything works!

packages/text-embeddings/chart/templates/deployment.yaml
```diff
@@ -25,6 +25,9 @@ spec:
       labels:
         {{- include "chart.selectorLabels" . | nindent 8 }}
     spec:
+      {{- if .Values.gpu.enabled }}
+      runtimeClassName: {{ .Values.gpu.runtimeClassName }}
```
YrrepNoj, line 29: Can you help me understand the _why_ we are adding this now? What has changed where this is now needed?

uds-bundles/latest/gpu/uds-config.yaml
```diff
@@ -1,13 +1,15 @@
...
 variables:
   text-embeddings:
-    gpu_limit: 0
+    gpu_enabled: true
+    gpu_limit: 0 # 0 to use cpu, 1+ to use gpu
 
   whisper:
-    gpu_limit: 0
+    gpu_enabled: true
+    gpu_limit: 0 # 0 to use cpu, 1+ to use gpu
```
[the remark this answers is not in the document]
justinthelaw: We can try that! I think that will work.

packages/text-embeddings/chart/templates/deployment.yaml
justinthelaw: [Underlying, custom k3s GPU image](https://github.com/justinthelaw/k3d-gpu-support) has changed to no longer inject a [hardcoded `config.toml.tmpl`](https://github.com/justinthelaw/k3d-gpu-support/blob/000422aa461512a7fcf927d4599bc8f6fa350b1a/config/config.toml.tmpl) that declares `nvidia` as the default runtime within the cluster's pods.

This upstream image change enables granular/explicit control of what runtime is used in the cluster. The 2 cases this change helps cover are as follows:

1. When a host system already has the container toolkit installed (only the k8s-device-plugin is installed via the underlying, custom k3s GPU image), we previously relied on the default/hardcoded `config.toml.tmpl` to force all pods to run on the `nvidia` runtime. This was fine because the runtime is provided by the host system's nvidia container toolkit installation. With the upstream changes to remove that hard-coded `config.toml.tmpl` (to better align with the default configurations of other k8s distros, like RKE2 (see next bullet for details on this)), this default `runtimeClass` injection no longer happens. This means the `runtimeClassName` must be specified in the pod deployment manifest. This all works great, HOWEVER, in the case where a CPU bundle is deployed and the `runtimeClassName: nvidia` exists in any manifest, and the host DOES NOT have CUDA drivers nor the NVIDIA container toolkit installed, an error will occur: `rpc error: runtime "nvidia" not found`. This is the main reason for exposing this variable and making it explicit and conditional in the manifest.
 
2. When the GPU operator is installed with `toolkit.enabled: true`, it automatically injects a containerd configuration that defaults the `runtimeClass` to `nvidia` for all pods. This change in our LFAI manifests will not conflict with this automatic injection, and the `nvidia` runtime can handle CPU or GPU workloads, as it is just `runc` under the hood with NVIDIA custom code on top. When the GPU operator is installed with `toolkit.enabled: false`, this change in our LFAI manifests helps the pods identify the `runtimeClass` to use for GPU workloads, as the default runtime is no longer declared/injected by the container toolkit automatically.

Beyond the above enhancements to compatibility in all possible installation cases, this change also sets us up for future refactors where the `runtimeClass` may be of a different type (ROCm, Metal, etc.) and not just `nvidia` for workload offloading/acceleration.

uds-bundles/latest/gpu/uds-bundle.yaml
```diff
@@ -36,13 +36,25 @@ packages:
...
     repository: ghcr.io/defenseunicorns/packages/leapfrogai/whisper
     # x-release-please-start-version
     ref: 0.9.1
     # x-release-please-end
+    overrides:
+      whisper-model:
+        whisper-model:
+          variables:
+            - name: gpu_enabled
+              path: gpu.enabled
```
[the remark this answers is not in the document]
justinthelaw: You are correct! We can just declare the variable within the `uds-config.yaml`, and add a comment saying this is a Zarf variable and not a UDS override variable. Similar to this: https://github.com/justinthelaw/uds-rke2/blob/67cbaebac73ec6e5aa2d13b4addebdac18ccdf3a/bundles/dev/local-path-core/uds-config.yaml

uds-bundles/latest/gpu/uds-config.yaml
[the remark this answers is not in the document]
gphorvath: The issue with this approach is that we won't be able to scale up the gpu limit in our deployments. By have `gpu_enabled: true` with `gpu_limit: 0` the container will install the GPU drivers necessary to run on GPU but then only run on CPU unless scaled up.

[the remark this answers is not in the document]
YrrepNoj: I'm not sure I follow. If we start with a `gpu_limit: 0` and want to 'scale up' to `gpu_limit: 2` then we still have to modify the deployment which will then recycle the pod. I don't see how that's different from just using `gpu_limit: 0` to also mean `gpu_enabled: false` since in both cases you will have to modify the deployment and cycle the pods to see any change.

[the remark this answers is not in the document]
gphorvath: I did try this, but when scaling up it doesn't append the runtimeClassName required.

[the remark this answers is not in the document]
gphorvath: I actually change the way I did this a bit. Feel free to comment on the new approach.

justinthelaw (dismissed):

uds-bundles/latest/cpu/uds-config.yaml
```diff
@@ -1,8 +1,11 @@
+# see individual zarf packaging configuration for more variables and variable descriptions
 variables:
   text-embeddings:
+    gpu_class_name: "" # Leave blank if nvidia runtimeClass is not present in cluster
     gpu_limit: 0
 
   whisper:
+    gpu_class_name: "" # Leave blank if nvidia runtimeClass is not present in cluster
     gpu_limit: 0
```
YrrepNoj, line 9: What would happen if we set `gpu_limit = 1` without setting the `gpu_class_name`?

justinthelaw: This would be a delivery engineer knowledge thing.

The container in the pod would fail. But helm template logic is not working when using injected Zarf variables - same issue we had with the leapfrogai-model-skeleton several months ago.

justinthelaw: Hm, maybe not that exact pattern, in my memory. We can retackle!

YrrepNoj: Made a [draft PR](https://github.com/defenseunicorns/leapfrogai/pull/821) that shows an example of what this logic would look like for this use case.

justinthelaw (approved):

packages/text-embeddings/zarf.yaml
```diff
@@ -16,6 +16,11 @@ variables:
     description: The GPU limit for the model inferencing.
     default: "0"
     pattern: "^[0-9]+$"
+  - name: GPU_CLASS_NAME
+    description: The GPU class name for the model inferencing. Leave blank for CPU-only.
+    default: ""
+    pattern: "^(nvidia)?$"
+    prompt: true
```
YrrepNoj, line 23: Can we remove the `prompt: true` for these? IMO setting this should be fairly niche.

gphorvath, line 23: don't prompt

gphorvath: done

YrrepNoj (approved):

PR #794, opened 2024-07-18
chore: Update defenseunicorns/zarf to zarf-dev/zarf

schristoff-du:


gphorvath:
This was addressed in a separate PR due to permission issues running the workflows: https://github.com/defenseunicorns/leapfrogai/pull/832

Issue #798, opened 2024-07-18
chore(api): Add load tests to pipeline


CollectiveUnicorn:
### Describe what should be investigated or refactored
Figure out if there's a way to meaningfully bring the load tests to the pipeline. If it is determined that they can be integrated meaningfully into the pipeline, add them to the pipeline.

PR #801, opened 2024-07-18
chore(api): Adds API load testing

CollectiveUnicorn:
* Adds load testing for the API that can be manually run from the test folder.


tests/load/README.md
CollectiveUnicorn, line 1: Create a make command to run this

CollectiveUnicorn: Make sure the issue explicitly calls out that it doesn't need to run with every push, maybe nightly. Possibly look into "test grid"?? Third party framework?

gphorvath: https://testgrid.io/

CollectiveUnicorn: Created make command and pipeline issue was updated with these notes: https://github.com/defenseunicorns/leapfrogai/issues/798

pyproject.toml
```diff
@@ -18,7 +18,9 @@ dependencies = [  # Dev dependencies needed for all of lfai
     "httpx",
     "ruff",
     "python-dotenv",
-    "pytest-asyncio"
+    "pytest-asyncio",
```
YrrepNoj, line 21: What do you think of using an optional dependency like `dev-dependencies` target for things that are only needed for tests?

tests/data/book.pdf
YrrepNoj, line 1: 5MB is starting to reach the point where we might want to avoid committing the binary if possible. What do you think about having the test download this file at runtime?

pyproject.toml
CollectiveUnicorn: I think it would be a good change to not have the dependencies mixed in together - I have created an issue for this here: https://github.com/defenseunicorns/leapfrogai/issues/809 - going to work on this in a separate PR as it seems like it will require a lot of changes unrelated to the load testing.

tests/load/loadtest.py
```diff
@@ -0,0 +1,246 @@
...
+        }
+        response = self.client.post("/openai/v1/vector_stores", json=payload)
+        if response.status_code == 200:
+            self.vector_store_id = response.json()["id"]
+            print(f"Created vector store ID: {self.vector_store_id}")
+
+    @task
+    def create_assistant(self):
+        payload = {
+            "model": "vllm",
```
gphorvath, line 113: Would be nice if model was configurable for cases with llama-cpp-python, etc.

tests/load/loadtest.py
```diff
@@ -0,0 +1,246 @@
...
+        self.client.get("/leapfrogai/v1/auth/list-api-keys")
+
+    @task
+    def test_openai_models(self):
+        self.client.get("/openai/v1/models")
+
+    @task
+    def test_chat_completions(self):
+        payload = {
+            "model": "vllm",
```
gphorvath, line 195: Would be nice if model was configurable for cases with llama-cpp-python, etc.

tests/load/README.md
```diff
@@ -0,0 +1,42 @@
...
+   ```bash
+   export API_URL="https://leapfrogai-api.uds.dev"
+   ```
+
+2. Set the API key:
+   ```bash
+   export API_KEY="<your-supabase-api-key-here>"
+   ```
+
+   **Note:** The API key should be your Supabase API key. For information on generating a key, please refer to the [Supabase README.md](../../packages/supabase/README.md). While an API key generated from the LeapfrogAI API endpoint can be used, it will cause the key generation load tests to fail.
```
gphorvath, line 23: I believe this is referring to generating a jwt. Might be a little confusing referring to it as `Supabase API key` as that could be construed as anon key or service key.

tests/load/loadtest.py
CollectiveUnicorn: I can update it, though the concurrency will be wonky if using llama-cpp-python.

tests/load/README.md
CollectiveUnicorn: Thanks, updated the verbage

tests/load/loadtest.py
CollectiveUnicorn: Should be configurable now

tests/load/loadtest.py
CollectiveUnicorn: Should be configurable now

gphorvath (dismissed):

YrrepNoj (approved):

PR #808, opened 2024-07-22
chore: split e2e.yaml into multiple workflows

jalling97:
This PR breaks out `e2e.yaml` into two separate workflows. Now that we have more Github runners, we can split the one large workflow into two smaller ones to speed up the process.


jalling97:
This PR is still in development but needs to be out of draft in order to test the e2e workflow. I'll follow up with another comment when it's ready.

jalling97:
Ready for review now

CollectiveUnicorn (approved):

justinthelaw (dismissed):

.github/workflows/e2e-2.yaml
```diff
@@ -1,4 +1,6 @@
-name: e2e
```
YrrepNoj, line 1: Can we split these out even more? It makes sense to me for us to have an e2e workflow per major package.
`e2e-vllm`
`e2e-text-embeddings`
`e2e-llama-cpp-python`
`e2e-whisper`

.github/workflows/e2e-llama-cpp-python.yaml
```diff
@@ -0,0 +1,117 @@
...
+      # Ignore LFAI-UI source code changes
+      - "!src/leapfrogai_ui/**"
+
+      # Ignore changes to unrelated packages
+      - "!packages/k3d-gpu/**"
+      - "!packages/repeater/**"
+      - "!packages/text-embeddings/**"
+      - "!packages/ui/**"
+      - "!packages/vllm/**"
+      - "!packages/whisper/**"
```
jalling97, line 46: I thought you wanted to avoid selectively running specific packages if we broke out the E2E tests this way? Just in case changes in one package impact how it interacts with another?

.github/workflows/e2e-playright.yaml
```diff
@@ -1,4 +1,6 @@
-name: e2e
+# End-to-end testing that deploys and tests Supabase, API, UI, and VLLM
+
+name: e2e-playright
```
jalling97, line 3:
```suggestion
name: e2e-playwright
```
Minor spelling in a few places and in the name of the workflow

.github/workflows/e2e-llama-cpp-python.yaml
YrrepNoj: Yessish. If the `api` or `supabase` packages get updated we want to make sure we are still testing the llama package since those changes might implicitly affect the usability of the llama package. However, us making updates to the whisper package would have no affect to the usage of the llama package.

jalling97: That's fair. I guess I was thinking once we get to running the API integration tests in CI but things are going to have to change for that anyway. Since it looks like this reduces the E2E test runtime from 60 minutes to 15-25 minutes, I'm down.

.github/workflows/e2e-llama-cpp-python.yaml
```diff
@@ -0,0 +1,117 @@
...
+  group: e2e-llama-cpp-python-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  e2e_llama:
+    runs-on: ai-ubuntu-big-boy-8-core
+    if: ${{ !github.event.pull_request.draft }}
+
+    steps:
+        - name: Checkout Repo
```
YrrepNoj, line 58: Look into making this an action that all the workflows can use to reduce code duplication.

.github/workflows/e2e-llama-cpp-python.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+        - name: Setup UDS Environment
+          uses: defenseunicorns/uds-common/.github/actions/setup@05f42bb3117b66ebef8c72ae050b34bce19385f5
+          with:
+            username: ${{ secrets.IRON_BANK_ROBOT_USERNAME }}
+            password: ${{ secrets.IRON_BANK_ROBOT_PASSWORD }}
+
+        - name: Create UDS Cluster
+          run: |
+            uds deploy k3d-core-slim-dev:0.22.2 --confirm
```
YrrepNoj, line 77: Use the uds common to deploy the cluster.

justinthelaw (approved):

gphorvath (approved):

PR #813, opened 2024-07-22
feat(ui): Flowbite header and sidebar

andrewrisse:
First PR for converting to Flowbite. 

* Adds Flowbite and TailwindCSS
* Refactors the header, sidebar, settings/user side drawers, and toast store
* Removes SASS Carbon preprocessor styling injects (ex implication: ```gap: layout.$spacing-03;``` **becomes** ```gap: 0.5rem;```)
* Fixes playwright tests so that they run locally again (broken with Playwright CI story)
* Updates tests so all unit and E2E tests are still passing despite mix of carbon and flowbite components

Deployed on LF-04
[LeapfrogAI - Chat - 22 July 2024 - Watch Video](https://www.loom.com/share/622b49a747cb4de3a27f2dc44cc18e84)


src/leapfrogai_ui/src/lib/components/LFHeader.test.ts
```diff
@@ -9,27 +9,31 @@ describe('LFHeader', () => {
     // We cannot test by visibility because carbon components svelte does not give us a way to grab the panel
```
CollectiveUnicorn, line 9: Is this comment still relevant given the removal of the `bx--header__action--active` code?

CollectiveUnicorn (changes requested):

src/leapfrogai_ui/src/lib/components/AssistantAvatar.svelte
```diff
@@ -200,10 +200,10 @@
     width: 3rem;
     height: 3rem;
     border-radius: 50%;
-    background-color: themes.$layer-active-03;
+    background-color: #393939;
```
BillyFigueroa, line 203: I would try to create css variables for some of these repeated reused values

src/leapfrogai_ui/src/lib/components/AssistantAvatar.svelte
```diff
@@ -236,19 +236,19 @@
...
   .edit-btns {
     display: flex;
-    gap: layout.$spacing-03;
+    gap: 0.5rem;
   }
 
   .error-box {
-    border: 2px solid $red-30;
-    color: $red-30;
+    border: 2px solid #ffb3b8;
```
BillyFigueroa, line 248: I would try to create css variables for some of these repeated reused values

src/leapfrogai_ui/src/lib/components/AssistantForm.svelte
```diff
@@ -237,14 +235,10 @@
...
     .title {
-      @include type.type-style('heading-05');
-    }
-
-    .cancel-modal {
-      :global(.bx--modal-container) {
-        position: absolute;
-        top: 25%;
-      }
+      font-size: 2rem;
```
BillyFigueroa, line 238: might help to actually create mixins or some scss related shortcut to get the same results

BillyFigueroa:
It looks like a good amount of the changes were un-doing a lot of the scss variables provided by carbon. I would likely recommend created some sort of js file you can export those types of values or doing something in scss/ postcss where you are creating your own utility classes, maybe even with the same name, to replace with and remove the problem of repetition as well as when you might need to make a chance in the future, you change it in one place

src/leapfrogai_ui/src/lib/components/LFHeader.test.ts
andrewrisse: it sure isn't

src/leapfrogai_ui/src/lib/components/AssistantAvatar.svelte
```diff
@@ -183,13 +183,13 @@
   .avatar-upload-container {
     display: flex;
     flex-direction: column;
-    gap: layout.$spacing-03;
+    gap: 0.5rem;
```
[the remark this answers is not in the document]
andrewrisse: This component will be moved over to Flowbite soon and will be using tailwind css. My understanding is that variables and re-usable styles are less of a thing, even frowned upon sometimes, with Tailwind. I've been using some Tailwind utilities though to have some re-usable stuff, but am limiting those use cases.

Do you have an example of using variables with tailwind? I'm curious if we can find a good pattern for that.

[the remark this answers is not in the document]
BillyFigueroa: > This component will be moved over to Flowbite soon and will be using tailwind css. My understanding is that variables and re-usable styles are less of a thing, even frowned upon sometimes, with Tailwind. I've been using some Tailwind utilities though to have some re-usable stuff, but am limiting those use cases.
> 
> Do you have an example of using variables with tailwind? I'm curious if we can find a good pattern for that.

Well, tailwind has a bunch of utility classes they have that are re-used. They are not using the actual hex codes for example in their color classes. You can create a utility class for spacing and color I believe. Might need to extend the tailwind config object and include them there

src/leapfrogai_ui/src/lib/components/AssistantAvatar.svelte
andrewrisse: expect this to change and likely go away when this component is moved to flowbite, for now this change is to remove the carbon sass pre-processing (caused a lot of performance issues)

src/leapfrogai_ui/src/lib/components/AssistantAvatar.svelte
andrewrisse: will address when this component is moved to flowbite, will change significantly

src/leapfrogai_ui/src/lib/components/AssistantForm.svelte
andrewrisse: will be using flowbite typography when this changes to flowbite

andrewrisse:
> It looks like a good amount of the changes were un-doing a lot of the scss variables provided by carbon. I would likely recommend created some sort of js file you can export those types of values or doing something in scss/ postcss where you are creating your own utility classes, maybe even with the same name, to replace with and remove the problem of repetition as well as when you might need to make a chance in the future, you change it in one place

Very valid point, but those changes are temporary (like changing tomorrow) as we move to flowbite. They just let the app compile while we remove the carbon sass precompile stuff.

BillyFigueroa:
> > It looks like a good amount of the changes were un-doing a lot of the scss variables provided by carbon. I would likely recommend created some sort of js file you can export those types of values or doing something in scss/ postcss where you are creating your own utility classes, maybe even with the same name, to replace with and remove the problem of repetition as well as when you might need to make a chance in the future, you change it in one place
> 
> 
> 
> Very valid point, but those changes are temporary (like changing tomorrow) as we move to flowbite. They just let the app compile while we remove the carbon sass precompile stuff. 

Copy 🙌🏽

CollectiveUnicorn (approved):

PR #821, opened 2024-07-23
example: Using helm templating to set runtime className

YrrepNoj:


packages/whisper/whisper-values.yaml
```diff
@@ -1,9 +1,6 @@
 image:
   tag: "###ZARF_CONST_IMAGE_VERSION###"
 
-gpu:
-  className: "###ZARF_VAR_GPU_CLASS_NAME###"
-
 resources:
   limits:
-    nvidia.com/gpu: "###ZARF_VAR_GPU_LIMIT###"
+    nvidia.com/gpu: ###ZARF_VAR_GPU_LIMIT###
```
YrrepNoj, line 6: Remove the quotes here because the string wouldn't work with the `gt` check. This could probably be converted to an int in the `deployment.yaml#28` to be more robust.

gphorvath:
@YrrepNoj can we close this? Up to you if you are happy with #787

YrrepNoj:
Closing as OBE. This was just to serve as an example :)

PR #824, opened 2024-07-23
feat: silent parallel make build all

gphorvath:
Adds silent build and deploy make targets that speed up the process of standing up LeapfrogAI by parallelizing builds/deployments.


Makefile
```diff
@@ -13,6 +13,9 @@ help: ## Display this help information
 		| sort | awk 'BEGIN {FS = ":.*?## "}; \
 		{printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
+silent-flags:
+    $(eval DOCKER_FLAGS := $(DOCKER_FLAGS) --quiet)
```
justinthelaw, line 17: Maybe also add `--progress plain` for the log files?

justinthelaw (approved):
1 minor comment, otherwise great idea!

Makefile
gphorvath: `--progress plain` and `quiet` are incompatible, so it really comes down to how little info we want to show. `quiet` is basically only going to show the most essential info (errors and final outputs) whereas `--progress plain` will give us more of the info logs along the way.

justinthelaw: For context, my main purpose in the past was to see if the llama.cpp compilation actually used the CUDA drivers to create the proper GPU-enabled `.whl`. Usually, that type of warning, or more like embedded error, doesn't show up unless you write the logs out . BUT, if our assumption for using a silent build is that it will all work and we don't need to read much of the logs.

gphorvath: maybe for debugging it's best to use `make build-llama-cpp-python`?

gphorvath:
Much faster from zero to working deployment
image

justinthelaw (approved):
Works beautifully! I think it is subjectively faster, although I failed to collect metrics on it...

PR #826, opened 2024-07-24
Fix Playwright tests running locally

andrewrisse:
Allows Playwright e2es to run locally again

### DIFF
--- a/src/leapfrogai_ui/playwright.config.ts
+++ b/src/leapfrogai_ui/playwright.config.ts
@@ -62,6 +62,8 @@ const devConfig: PlaywrightTestConfig = {
     port: 4173,
     stderr: 'pipe'
   },
+  testDir: 'tests',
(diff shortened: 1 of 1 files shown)


gphorvath (changes requested):
Now that the UI is encapsulated in the e2e test, we should modify the e2e and e2e-shim workflows to stop ignoring changes in the ui directories so that changes like this trigger the e2e.

https://github.com/defenseunicorns/leapfrogai/blob/134c693fdd5b8e71d405bc51d6567285031df1a8/.github/workflows/e2e.yaml#L35

https://github.com/defenseunicorns/leapfrogai/blob/134c693fdd5b8e71d405bc51d6567285031df1a8/.github/workflows/e2e-shim.yaml#L26

gphorvath (approved):

jalling97 (approved):

PR #830, opened 2024-07-24
feat(api): adds migration for tracking vector store indexing status

CollectiveUnicorn:
* Enables Supabase realtime for the `vector_store_file` table
* Adds new Python dependency to test supabase-realtime
* Status is currently being set during indexing so listening to this table should provide status updates on the file indexing process.

https://github.com/defenseunicorns/leapfrogai/blob/30b8ba2f0b9b211c9f84a452895c08f14865d9c8/src/leapfrogai_api/backend/rag/index.py#L41


justinthelaw:
Can we modify the tests, and also create a new test for ensuring this works?

CollectiveUnicorn:
Yeah I'll create a test next week for this

justinthelaw (approved):

justinthelaw (approved):

tests/e2e/test_supabase.py
```diff
@@ -12,9 +31,149 @@
...
+        s = Socket(URL)
+        s.connect()
+
+        # Set channel to listen for changes to the vector_store_file table
+        channel_1 = s.set_channel("realtime:public:vector_store_file")
+        # Listen for all events on the channel ex: INSERT, UPDATE, DELETE
+        channel_1.join().on("*", postgres_changes_callback)
+
+        # Start listening
+        s.listen()
```
alekst23, line 172: Given that this is a strongly blocking operation, as you mentioned, will it have an impact on our API stack if we deploy this as a service endpoint? For example, will a single user requesting the endpoint for file status end up blocking the whole server?

CollectiveUnicorn: No, this is something that only exists in the test, the frontend can subscribe to these changes without blocking and the library isn't used in the backend code. The backend code is only db changes.

jalling97 (approved):
Just needs to be updated with `main` and everything looks good to me!

packages/api/supabase/migrations/20240724103100_v0.9.2_indexing_status.sql
```diff
@@ -0,0 +1,24 @@
+-- Update the vector_store_file table to add an updated_at column
+alter table vector_store_file add column updated_at bigint default extract(epoch from now()) not null;
```
alekst23, line 2: Is there a benefit to using `bigint` over `timestamp`? Currently, this will field will have precision in seconds. It would be easier to search and sort this data later if we use the `timestamp` type with millisecond precision, as a standard for timestamp fields:
```
ALTER TABLE vector_store_file ADD COLUMN updated_at timestamp(3) DEFAULT CURRENT_TIMESTAMP NOT NULL;
```

packages/api/supabase/migrations/20240724103100_v0.9.2_indexing_status.sql
```diff
@@ -0,0 +1,24 @@
...
+
+-- Add an index on user_id for faster queries
+CREATE INDEX idx_vector_store_file_user_id ON vector_store_file(user_id);
+
+-- Create a function to update the updated_at column
+CREATE OR REPLACE FUNCTION update_modified_column()
+RETURNS TRIGGER AS $$
+BEGIN
+    NEW.updated_at = extract(epoch from now());
```
alekst23, line 11: If we use `timestamp` type, this line becomes:
```
NEW.updated_at = CURRENT_TIMESTAMP;
```

pyproject.toml
```diff
@@ -27,6 +27,8 @@ dev = [
     "requests-toolbelt",
     "pytest",
     "huggingface_hub[cli,hf_transfer]",
+    "fastapi",
+    "supabase == 2.6.0"
```
gphorvath, line 31: Curious what the rationale is for adding these here, versus the api `pyproject.toml`

tests/e2e/test_supabase.py
```diff
@@ -12,9 +31,149 @@
...
+
+        # Schedule postgres_db_changes to run after 5 seconds
+        threading.Timer(5.0, run_postgres_db_changes).start()
+
+        # Set a timeout of 10 seconds
+        timeout_timer = threading.Timer(10.0, timeout_handler)
+        timeout_timer.start()
+
+        # Listening socket
+        URL = f"wss://supabase-kong.uds.dev/realtime/v1/websocket?apikey={SERVICE_KEY}&vsn=1.0.0"
```
gphorvath, line 162: Maybe add a comment why we need the service key for testing and that RLS will be observed by the app in prod.

packages/supabase/chart/templates/suapbase-realtime-secret.yaml
```diff
@@ -0,0 +1,18 @@
+{{- $dbEncKey := randAlphaNum 16 }} # This needs to be exactly 16 characters
```
jalling97, line 1: presumably this file should be named `supabase-realtime-secret.yaml` and not `suapbase-realtime-secret.yaml`

packages/api/supabase/migrations/20240724103100_v0.9.2_indexing_status.sql
gphorvath, line 1: the version has updated to v0.10.x since starting this PR, so this should be renamed

pyproject.toml
CollectiveUnicorn: They are needed for the tests to run, the workflow only adds dev deps from the at the root of the project

packages/supabase/chart/templates/suapbase-realtime-secret.yaml
CollectiveUnicorn: Yeah this was updated in another PR so this file is not needed

packages/api/supabase/migrations/20240724103100_v0.9.2_indexing_status.sql
CollectiveUnicorn: Bumped it to 0.11.0

tests/e2e/test_supabase.py
CollectiveUnicorn: Added comment

gphorvath (approved):
Tested locally and everything seems to be working and all my comments have been addressed.

packages/api/supabase/migrations/20240724103100_v0.9.2_indexing_status.sql
CollectiveUnicorn: Updated to timestamp

packages/api/supabase/migrations/20240724103100_v0.9.2_indexing_status.sql
CollectiveUnicorn: Updated to suggested timestamp. There is no benefit in this case, we are using bigints a few places due to their mapping with openai but this isn't one of those places.

packages/api/supabase/migrations/20240724103100_v0.9.2_indexing_status.sql
YrrepNoj: @CollectiveUnicorn 0.11.0 was the last release. This file should either be 0.11.1 or 0.12.0.

I liked the idea of knowing what release version each migration file was introduced but keeping this maintained is starting to get pretty gross and is starting to feel like it's going to be hard to maintain. :/

gphorvath: It's probably not going to be easy to nail the version, because somebody could merge a PR that introduces a breaking change. It might be just as well to go by the date in the timestamp to see when the next release tag was cut and make sure to update the date shortly before merging to main.

gphorvath (approved):

packages/api/supabase/migrations/20240724103100_v0.9.2_indexing_status.sql
CollectiveUnicorn: I've bumped it to 0.11.1, I feel like time is probably as good as we're going to get. But this should adhere to our current pattern for now.

gphorvath (approved):

tests/e2e/test_supabase.py
```diff
@@ -12,9 +31,151 @@
...
+            status="completed",
+            usage_bytes=0,
+        )
+
+        await CRUDVectorStore(client).create(vector_store)
+
+        vector_store_file = VectorStoreFile(
+            id=upload_file_id,
+            vector_store_id=vector_store.id,
+            created_at=int(time.time()),
```
alekst23, line 85: The table has a default value on created_at:
`created_at bigint default extract(epoch FROM NOW()) not null,`
Do we still need to include it in our setup code?

gphorvath: We typically set this to zero to make the VectorStoreFile constructor happy.

CollectiveUnicorn: You're right we don't explicitly need to do this, but for what is effectively the same functionality in a test I don't see much value in changing this.

CollectiveUnicorn: The test is aimed at verifying that events are generated by making changes to the db, whether the created_at time is autogenerated by the db or explicitly defined leads to the same result.

packages/api/supabase/migrations/20240724103100_v0.9.2_indexing_status.sql
gphorvath: I've been sleeping on this an I think we should continue to use epoch time with higher resolution. I don't think timestamp is timezone safe. Let me know if I'm off base here.

CollectiveUnicorn: You are right, we can specify a timestamp without a timezone and use `LOCALTIMESTAMP` to do so without the timezone, but I am a little wary to start assuming timestamp conversions. I think unix time is safer to use.

CollectiveUnicorn: I will test the conversion though, if I don't see any issues I'll probably do that. As ms for unix time in pg seems non-standard.

CollectiveUnicorn: Ok I made some additional modifications, will be sticking with mostly the same as what we have now but I've explicitly set the time zone to utc and not specified a precision level for the moment. This is based on the recommendations here: https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp_.28without_time_zone.29 we can still truncate it to millisecond precision if we want but I'm not sure its worth the extra operation.

tests/e2e/test_supabase.py
CollectiveUnicorn: Changed this to 0 to match how we have it elsewhere.

gphorvath (dismissed):

tests/e2e/test_supabase.py
```diff
@@ -12,9 +30,151 @@
...
+            supabase_key=ANON_KEY,
+            supabase_url="https://supabase-kong.uds.dev",
+        )
+        await client.auth.set_session(access_token=access_token, refresh_token="dummy")
+
+        upload_file_id = await upload_file(client)
+        assert upload_file_id is not None, "Failed to upload file"
+
+        vector_store = VectorStore(
+            id=str(uuid.uuid4()),
```
gphorvath, line 64: This has a worrying smell. Why do we need to set a uuid? This should be handled by the DB and return in the vector store object returned by the db.

tests/e2e/test_supabase.py
```diff
@@ -12,9 +30,151 @@
...
+        )
+
+        if event_information_match and all_records_match:
+            raise TestCompleteException("Test completed successfully")
+
+    async def upload_file(client: AsyncClient) -> str:
+        """
+        This function is responsible for uploading a file to the file bucket.
+        """
+        id_ = str(uuid.uuid4())
```
gphorvath, line 118: This has a worrying smell. Why do we need to set a uuid? This should be handled by the DB and come back in the file_object returned by the crud call.

tests/e2e/test_supabase.py
CollectiveUnicorn: Yeah it is odd, I'll look into it more

tests/e2e/test_supabase.py
CollectiveUnicorn: The id was indirectly being used in multiple places, more was required than simply replacing the value to fix the issue.

tests/e2e/test_supabase.py
CollectiveUnicorn: There were some indirect references to the id that assumed they has already been created on the `vector_store` object. By swapping this id with `""` those references broke. So I needed to also update the indirect references before it would work as expected.

gphorvath (dismissed):
Looks good to me once you resolve the conflict with main

justinthelaw (approved):

jalling97 (approved):
You may need to update this with main, if that dismisses the review I'll check it again :+1:

gphorvath (approved):

justinthelaw (approved):

PR #833, opened 2024-07-25
fix(ui): playwright login without keycloak

andrewrisse:
The Supabase Auth UI component is no longer being maintained and the auth flow when not using Keycloak was broken.

Playwright in the workflow pipeline currently cannot use Keycloak, so login with Keycloak turned off is necessary (direct login with Supabase).

This PR fixes the login flow. The login page with keycloak off will be re-styled separately with the Flowbite refactor.

The only test for this flow is an e2e with Keycloak turned off. It does not test the error cases or email confirmation, we can add those tests later if we want, but moving to Authservice may make all of this obsolete. For speed, it was decided not to fully test this flow, and just to release this patch to unblock the pipeline.


jalling97:
Looks good; mind if I update your PR so that the E2E tests run on UI changes? Just to make sure this is passing in CI now. Or you can; just need to remove a couple of lines in `e2e.yaml` and `e2e-shim.yaml`.

andrewrisse:
> Looks good; mind if I update your PR so that the E2E tests run on UI changes? Just to make sure this is passing in CI now. Or you can; just need to remove a couple of lines in `e2e.yaml` and `e2e-shim.yaml`.

please do! I was trying to figure out why it didn't run

jalling97 (approved):

jalling97 (approved):

CollectiveUnicorn (dismissed):

jalling97 (approved):

justinthelaw (approved):

PR #840, opened 2024-07-25
chore(whisper): Pass through variables down to whisper

CollectiveUnicorn:
Ensures that all variables are passed down to the actual whisper model
* `language, temperature, prompt`

### DIFF
--- a/packages/whisper/main.py
+++ b/packages/whisper/main.py
@@ -27,7 +27,7 @@ def make_transcribe_request(filename, task, language, temperature, prompt):
             logger.error(f"Task {task} is not supported")
             return {"text": ""}
     if language:
-        if language in WhisperModel.supported_languages:
+        if language in WhisperModel.supported_languages():
             kwargs["language"] = language
         else:
             logger.error(f"Language {language} is not supported")
--- a/packages/whisper/main.py
+++ b/packages/whisper/main.py
@@ -27,7 +27,7 @@ def make_transcribe_request(filename, task, language, temperature, prompt):
             logger.error(f"Task {task} is not supported")
             return {"text": ""}
     if language:
-        if language in WhisperModel.supported_languages():
+        if language in model.supported_languages():
             kwargs["language"] = language
         else:
             logger.error(f"Language {language} is not supported")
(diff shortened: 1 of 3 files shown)


justinthelaw (dismissed):

gphorvath (dismissed):

packages/whisper/main.py
```diff
@@ -18,10 +18,21 @@ def make_transcribe_request(filename, task, language, temperature, prompt):
     device = "cuda" if GPU_ENABLED else "cpu"
     model = WhisperModel(model_path, device=device, compute_type="float32")
 
-    segments, info = model.transcribe(filename, task=task, beam_size=5)
+    # Prepare kwargs with non-None values
+    kwargs = {}
+    if task is not None:
+        kwargs["task"] = task
```
gphorvath, line 24: should we validate that task is `transcribe` or `translate` or just leave that up to faster-whisper?

packages/whisper/main.py
gphorvath, line 1: Is this a good time to upgrade the version of faster-whisper? I noticed we are on `0.1.0` which is from November 2023. Version 1.0.3 (July 2024) seem to work locally for me with these changes.

packages/whisper/main.py
```diff
@@ -18,10 +18,21 @@ def make_transcribe_request(filename, task, language, temperature, prompt):
...
+    if language is not None:
+        kwargs["language"] = language
```
gphorvath, line 26: faster-whisper has `supported_languages() -> List[str]` that we could use to validate an incoming language

packages/whisper/main.py
CollectiveUnicorn: Yep, bumping the version

packages/whisper/main.py
CollectiveUnicorn: Added a check for this and made it default to automatically recognizing the language if that fails

packages/whisper/main.py
CollectiveUnicorn: I've added checks here now

justinthelaw (approved):

PR #852, opened 2024-07-29
fix(api): fix indexing files with api key auth

gphorvath:


CollectiveUnicorn (dismissed):

justinthelaw:
Are the llama-cpp-python changes in this PR part of getting the file indexing to work locally?

justinthelaw (dismissed):

packages/api/supabase/migrations/20240729193626_api_keys_storage_objects.sql
```diff
@@ -0,0 +1,11 @@
+create policy "Individuals can CRUD storage.objects via API key."
```
YrrepNoj, line 1: Leaving a note to myself to re-read this again tonight/tomorrow with fresh eyes to make sure I understand this. <3

packages/llama-cpp-python/Makefile
```diff
@@ -1,7 +1,20 @@
-install:
```
YrrepNoj, line 1: I agree with @justinthelaw. If the changes to `packages/llama-cpp-python` are not relevant to fixing the issue with the file indexing I would rather not include them in this PR. Especially because this now doesn't match the Make targets names in the other package directories.

src/leapfrogai_api/backend/rag/leapfrogai_embeddings.py
```diff
@@ -44,7 +45,10 @@ async def aembed_query(self, text: str) -> list[float]:
 
         return list_of_embeddings[0]
 
-    async def _get_model(self, model_name: str = os.getenv("DEFAULT_EMBEDDINGS_MODEL")):
+    async def _get_model(
+        self,
+        model_name: str = os.getenv("DEFAULT_EMBEDDINGS_MODEL") or "text-embeddings",
```
YrrepNoj, line 50: Instead of using `or` we should use the optional param in `os.getenv( ... )`

```suggestion
        model_name: str = os.getenv("DEFAULT_EMBEDDINGS_MODEL", "text-embeddings)
```

YrrepNoj (changes requested):
A couple small requests/changes.

YrrepNoj (approved):

justinthelaw (approved):

Issue #835, opened 2024-07-25
feat(vllm): upgrade vllm and expose more params for bfloat16 quant compatibility


justinthelaw:
### Describe what should be investigated or refactored

vLLM is currently not compatible with all GPTQ BFLOAT16 quanitzed models due to the dependency version (0.4.2). This needs to be upgraded to the next patch version (0.4.3), or just completely upgraded to the next minor version (0.5.2).

The following test model should work if this issue has been fixed (fits on RTX 4060 - 4090): https://huggingface.co/TheBloke/phi-2-orange-GPTQ/blob/main/config.json

### Links to any relevant code

Example model that wouldn't work, but should: https://huggingface.co/TheBloke/phi-2-orange-GPTQ/blob/main/config.json

Issue related to the vLLM GPTQ BFLOAT16 PR: https://github.com/vllm-project/vllm/issues/2149

### Additional context

This issue was confirmed when deploying [Nous-Hermes-2-8x7b-DPO-GPTQ (8bit, 128g group size and Act Order)](https://huggingface.co/TheBloke/Nous-Hermes-2-Mixtral-8x7B-SFT-GPTQ/tree/gptq-8bit-128g-actorder_True) to an H100 GPU. Changing the `config.json` to `float16`, despite imprecision, allows the model to be inferenced by vLLM.

justinthelaw:
https://github.com/vllm-project/vllm/issues/2422

justinthelaw:
Additionally, the configuration we pass to vLLM should not include quantization, as that prevents marlin_gptq quantization which uses a different algorithm to perform faster inferencing and less memory usage.

Also, `trust_remote_code` refers to the code downloaded as part of the model download, so this can be safely turned on as long as we review the extra Python scripts downloaded as part of the model download. These scripts usually just tell vLLM how to configure itself for inferencing the model architecture (e.g., Phi-3 GPTQ).

justinthelaw:
![Screenshot 2024-07-30 103756](https://github.com/user-attachments/assets/5ad653b6-9bf1-43cf-8a3c-c7aa21e46b29)
![Screenshot 2024-07-30 103906](https://github.com/user-attachments/assets/2da22a0d-13d3-4894-a770-739f5aea809c)

Above screenshots comparing (generally) Phi-3-mini-128k-instruct outperforming all other Mistral-7b-instruct variants.

[...]

Issue #623, opened 2024-06-13
feat(api, backend) Implement Model Directory


barronstone:
### Overview

LeapfrogAI should implement a model directory to store and serve models. This would significantly reduce the size of LeapfrogAI Zarf packages, provide users with more model options, and enable the backends to dynamically swap out models for different tasks.

### Background

Currently, model parameters are directly “baked in” to backend Zarf packages. This makes LeapfrogAI packages extra large, often exceeding 8GB, which causes automated deployment issues that require manual intervention to work around. Additionally, each backend, such as `vllm`, is only packaged with a single model. If a user wants access to multiple LLMs for different tasks (e.g., one model for chat and another for coding assistance) then they would need two `vllm` packages, one for each model. That current approach makes it cumbersome and impractical for users to switch between models.

Externally, we are hearing demand for users to be able to select from a collection of models to use from organizations like Platform One. This is a common and expected feature that is available in most modern AI chat interfaces. Internally, for the LeapfrogAI team to efficiently perform evaluations on a variety of models, we need a way to quickly and efficiently swap out models. Incorporating a model directory into LeapfrogAI could solve these challenges - decoupling model parameters from Zarf packages and enabling users to dynamically select which models to use from those available in the model directory.

### Goals

 - Models are no longer directly “baked in” to the backend Zarf packages
 - Multiple models can be included in the UDS bundle for initial air-gapped deployment
 - The model directory can store and serve one or more models for each of the following types:
   - Text-to-text (LLM)
   - Text-to-vector (Embeddings)
   - Speech-to-text (Whisper)
 - System administrators can add new models to the model directory of an existing LeapfrogAI deployment in an air-gapped manner (No Internet connection required)
 - Users can select from available LLM models for an Assistant to use via the GUI and the API

### User Stories

**As a** delivery engineer deploying LeapfrogAI
**I want** the models to be separate from the backend Zarf packages
**So that** I can easily choose which models to deploy without having to rebuild packages, and 
**So that** the packages are smaller and therefore do not require manual steps to push them

**As a** LeapfrogAI end user
**I want** to be able to select from multiple LLMs
**So that** I can choose the best model to use for a specific task

### Acceptance Criteria - TODO

**Given** [a state]  
**When** [an action is taken]  
**Then** [something happens]

### Additional context
In-work technical design doc in LeapfrogAI Coda: https://coda.io/d/_dGmk3eNjmm8/Model-Directory_suuoJYJF

Issue #717, opened 2024-07-02
(spike) Determine Open Source LLMs for Initial Evaluations


jalling97:
## Description
As part of the deliverables for an MVP Evals framework, we will need a short list of LLMs we are evaluating on as part of LFAI. The models chosen should fit the following criteria (with room for potential exceptions):
- Fewer than 5 total models (excess evals at the beginning is information overload, requires extra time, and may not be valuable out of the gate)
- Models should be foundational (avoid hyper-specific fine-tuned models)
- Models should be Apache-2.0 (or more permissible)
- Models should be able to fit on 12-16Gb VRAM

## Relevant Links
Inspiration can be found on the [HuggingFace Open LLM Leaderboard](https://huggingface.co/spaces/open-llm-leaderboard/open_llm_leaderboard)

jalling97:
Since we implement [Synthia-7B](https://huggingface.co/TheBloke/SynthIA-7B-v2.0-GGUF) as our default model, it will be included in the list of initial evaluations.

We have also done our own quantization of a fine-tuned [Mistral-7B](https://huggingface.co/defenseunicorns/Hermes-2-Pro-Mistral-7B-4bit-32g-GPTQ), so this should be another model used in the initial evaluations.

jalling97:
Potential models for consideration:
- [Phi-3-small-instruct](https://huggingface.co/microsoft/Phi-3-small-128k-instruct), in either the 8k or 128k context (also available for ONNX)
- [Mistral-NeMo](https://huggingface.co/mistralai/Mistral-Nemo-Instruct-2407) (currently debating on size constraint concerns)

jalling97:
Quick update to balance scope and time; the following models will be used:
- SynthIA-7B (default model for now)
- Hermes2 (Defense Unicorns quantization)
- LLama3.1-8B (pending license approval)
    - If not available, Phi-3-small (8k context)

GPT-4o will also be used as a point of comparison in the results.

jxtngx:
Hi, @jalling97! Have you seen NVIDIA's distilled Llama 3.1 8B models?

1. https://huggingface.co/nvidia/Llama-3.1-Minitron-4B-Width-Base
2. https://huggingface.co/nvidia/Llama-3.1-Minitron-4B-Depth-Base

Here's the paper on model distillation, if interested: https://arxiv.org/abs/2408.11796

jalling97:
[...]

PR #854, opened 2024-07-30
feat(vllm)!: upgrade vllm backend and refactor deployment

justinthelaw:
## OVERVIEW

See https://github.com/defenseunicorns/leapfrogai/issues/835 for more details on rationale and findings. Also related to https://github.com/defenseunicorns/leapfrogai/issues/623.

**IMPORTANT NOTE:** there are still ongoing [AsyncLLMEngineDead](https://github.com/vllm-project/vllm/issues/5901#issuecomment-2290447888) and [RoPE scaling + Ray](https://github.com/vllm-project/vllm/issues/6607) issues upstream that may prevent us from upgrading past `0.4.x`.

## BREAKING CHANGES:

- moves all ENV specific to LeapfrogAI SDK to a ConfigMap using `volumeMount` for runtime injection and modification
  - in local dev, this is defined via `config.yaml`
- moves all ENV specific to vLLM to a ConfigMap, using `envFrom` for runtime injection and modification
  - in local dev, this is defined via `.env`
- `ZARF_CONFIG` is used to define create-time and deploy-time variables for (e.g., `MODEL_REPO_ID`, `ENFORCE_EAGER`)
  - updates Make targets and workflows with new `ZARF_CONFIG` variable
  - updates UDS bundles with new Zarf deployment variable overrides
  - allows delivery engineer's declarative definition of the backend configs and model
- re-introduces LFAI SDK `config.yaml` configuration method for local development and testing
- MUST upgrade API and backends together due to `FinishReason` proto change

## CHANGES:

- updates docs for running vLLM locally, in docker, and in-cluster
- update to Python 3.11.9 to align with Registry1 base image
- upgrades vLLM from 0.4.2 to 0.4.3 for:
    - adds BFloat16 quantized model support (the "why is this important" [here](https://www.cerebras.net/machine-learning/to-bfloat-or-not-to-bfloat-that-is-the-question/))
- exposes more backend configurations for Zarf build (see `packages/vllm/README.md`)
  - adds full set of `QUANTIZATION` options to existing configuration field
  - exposes everything via a Zarf variable and the values files
- removes default vLLM engine configurations from the Dockerfile (no duplicates, hardcoding)
- fixes issue where request params are not used in generation of the response (e.g., temperature)
  - uses LFAI SDK-received request object for inferencing params
  - gracefully handles vLLM versus SDK param names/differences for generation

## ADDITIONAL CONTEXT

The default model is still Synthia-7b, until #976 is resolved. The below description is only being kept for future context:

> The new model option, `defenseunicorns/Hermes-2-Pro-Mistral-7B-4bit-32g`, takes ~4.16Gb to load on to RAM or vRAM. At 6Gb of vRAM, the `max_context_length` variable has to be reduced to ~400 tokens; at 8Gb vRAM, ~10K tokens; and at 12Gb of vRAM, ~15K tokens. All cases require the `max_gpu_utilization` variable to be `0.99` in order to max out the KV cache size for the context length that is to be reserved at inference.
> 
> To achieve the maximum context length (~32K tokens) of the model, 16Gb of vRAM is required.
> 
> Reaching the defined `max_context_length` during a completion or chatting will result in the activation of vLLM's automatic sliding window handler, which drops quality of the final responses significantly.


justinthelaw:
For PR review, I recommend deploying testing out different build arguments (see the Dockerfile), to include swapping out the model and default backend configurations, to make sure things work with the new vLLM engine.

justinthelaw:
See this issue on CPU offloading: https://github.com/vllm-project/vllm/issues/6952

The offloading works... but the actual inferencing runs into issues when certain parts of the code are looking for a model weight that isn't where it expects.

justinthelaw:
Converting back to draft - some odd engine behavior / interaction with quantized Phi-3. Investigating...

justinthelaw:
Screenshot 2024-07-31 165127

The LLM responds correctly; however, the engine runs into some blocking process that causes a timeout on subsequent responses when the conversation reaches 2-5 messages long.

packages/vllm/.env.example
```diff
@@ -1,13 +1,17 @@
...
 export LAI_MODEL_SOURCE=".model/"
 export LAI_MAX_CONTEXT_LENGTH=32768
-export LAI_STOP_TOKENS='["</s>","<|endoftext|>","<|im_end|>"]'
-export LAI_PROMPT_FORMAT_CHAT_SYSTEM="SYSTEM: {}\n"
-export LAI_PROMPT_FORMAT_CHAT_ASSISTANT="ASSISTANT: {}\n"
-export LAI_PROMPT_FORMAT_CHAT_USER="USER: {}\n"
+export LAI_STOP_TOKENS='["</s>"]'
+export LAI_PROMPT_FORMAT_CHAT_SYSTEM="<|system|>\n{}<|end|>\n"
+export LAI_PROMPT_FORMAT_CHAT_USER="<|user|>\n{}<|end|>\n"
+export LAI_PROMPT_FORMAT_CHAT_ASSISTANT="<|assistant|>\n{}<|end|>\n"
```
justinthelaw, line 11: Prompt template should be ChatML format.

For activating the function calling and JSON response modes in the LLM, additional prompt formats, code and documentation should be provided.

packages/vllm/.env.example
```diff
@@ -1,13 +1,17 @@
 export LAI_HF_HUB_ENABLE_HF_TRANSFER="1"
-export LAI_REPO_ID="TheBloke/Synthia-7B-v2.0-GPTQ"
-export LAI_REVISION="gptq-4bit-32g-actorder_True"
-export LAI_QUANTIZATION="gptq"
+export LAI_REPO_ID="justinthelaw/Hermes-2-Pro-Mistral-7B-4bit-32g"
```
gphorvath, line 2: We should probably use models published under a Defense Unicorns organization

justinthelaw: Yes! Will transfer and create a repository once someone else confirms this is working too.

Also, I need to try to run a Locust test against this model as compared to Synthia-7b quantized.

justinthelaw:
https://huggingface.co/spaces/mlabonne/Yet_Another_LLM_Leaderboard - for checking Hermes-2-Pro-Mistral-7b's current standings

packages/vllm/src/config.py
```diff
@@ -5,39 +5,85 @@
...
+    enforce_eager: bool = Field(
+        default=True,
+        title="Enable Eager Mode",
+        description="Enable eager mode to start token generation immediately after prompt processing."
+        "Potentially reduces initial latency at the cost of slightly higher memory usage."
+        "Should be set to False in production environments with higher GPU memory.",
+        examples=[True, False],
+    )
+    gpu_memory_utilization: float = Field(
+        default=0.99,
```
alekst23, line 24: It looks like we are already setting an env var above for this purpose:
`LAI_GPU_MEMORY_UTILIZATION=0.9`
And similarly for other fields below. 

Should we set these default values from env to avoid duplicate definitions and conflicts?

From this config it's unclear which will take precedent.

justinthelaw: Ah yeah, that's true. We can keep the defaults within the Dockerfile-level ENV, since those are where they should be set.

justinthelaw:
Blocked by the completion and results of #717

justinthelaw:
Related to #623  - using hot-swappable `ConfigMap` and introducing Zarf variables to influence model download and configuration

justinthelaw:
Blocked by https://github.com/defenseunicorns/leapfrogai/pull/1038

justinthelaw:
Blocked until #1085 and #1038 are merged into main.

packages/vllm/Dockerfile
```diff
@@ -6,7 +6,7 @@ FROM nvidia/cuda:12.2.2-devel-ubuntu22.04 AS builder
 # set SDK location
 # set the pyenv and Python versions
```
justinthelaw, line 7: Add a Makefile command for cases where ONLY the Docker container is built and run, and it needs the .env and configmap to be bound as a volume and envfile.

CollectiveUnicorn (approved):

justinthelaw:
Blocked by IronBank issue [#22](https://repo1.dso.mil/dsop/opensource/defenseunicorns/leapfrogai/api/-/issues/22)

justinthelaw:
Awaiting completion of #1101

justinthelaw:
Blocked by https://repo1.dso.mil/dsop/opensource/defenseunicorns/leapfrogai/api/-/merge_requests/39

CollectiveUnicorn (dismissed):

tests/utils/client.py
```diff
@@ -1,9 +1,99 @@
...
+    """
+    headers = {
+        "apikey": f"{anon_key}",
+        "Authorization": f"Bearer {anon_key}",
+        "Content-Type": "application/json",
+    }
+
+    try:
+        requests.post(
+            url="https://supabase-kong.uds.dev/auth/v1/signup",
```
gphorvath, line 45: minor nit - this could be exposed via the `SUPABASE_URL` env

justinthelaw: Ah true! I'll integrate that into both branches. vLLM takes from the registry1 branch #1101

CollectiveUnicorn (dismissed):

tests/utils/client.py
```diff
@@ -1,9 +1,113 @@
...
+
+
+def get_supabase_url() -> str:
+    """Get the URL for Supabase.
+
+    Returns:
+        str: The URL for Supabase. (default: "https://supabase-kong.uds.dev/auth/v1")
+    """
+
+    return os.getenv("SUPABASE_URL", "https://supabase-kong.uds.dev/auth/v1")
```
gphorvath, line 24: I think we want this to return the base URL by default `https://supabase-kong.uds.dev` and append `auth/v1` for the create_test_user endpoints - just in case there are other tests that hit supabase that could consume this method

justinthelaw: Ah yeah, I made the change in registry1 but didn't pull-in here. Will do!

src/leapfrogai_api/typedef/completion/completion_types.py
```diff
@@ -11,7 +11,11 @@ class FinishReason(Enum):
     STOP = 1  # Maps to "stop"
     LENGTH = 2  # Maps to "length"
 
-    def to_string(self) -> str | None:
+    def __init__(self, value):
+        if value not in {0, 1, 2}:
```
gphorvath, line 15: `if value not in {FinishReason.NONE, FinishReason.STOP, FinishReason.LENGTH}:`

justinthelaw: I actually am going to refactor this based on your suggestion, but in a different way after re-looking at the base enum class in Python.

gphorvath (approved):

packages/vllm/README.md
```diff
@@ -39,6 +47,14 @@ uds zarf package deploy packages/vllm/zarf-package-vllm-*-dev.tar.zst --confirm
 
 ### Local Development
 
+The [config.yaml](./config.yaml) and [.env.example](./.env.example) must be modified if the model has changed away from the default.
+
+Create a `.env` file based on the [.env.example](./.env.example):
```
jalling97, line 52: To clarify, `config.yaml` and `.env.example` are only used for local dev, right? 

If that assumption is correct, can we add a quick note here stating that's the case? I know it's implied, but explicitly stating so (unless I'm missing where that is) might help people from confusing things

justinthelaw: Ah you are right - I should be more specific. In this README it is implicit, and the Docker Make target also imply it, but it definitely isn't clear now that I am reading this over again.

jalling97 (approved):
Glad to see the process for updating models is simplified (both locally and in deployments) and the docs are clear as to how it works! Can confirm that vllm builds and deploys to a local cluster without issue. It's a shame about the issues with newer version of vLLM, there are some features that would be nice to utilize. Great stuff!

gphorvath (approved):

PR #855, opened 2024-07-30
chore(supabase): fix supabase-realtime

CollectiveUnicorn:
These changes makes `supabase-realtime` functional both on the admin dashboard and for external connections:
* Adds missing mandatory environment variables
* Adds route for supabase realtime's API in addition to the one that exists for websocket connections
* Runs built in supabase realtime db seed command on pod startup to setup tenant
* Updates all variables to use the name `supabase-realtime` as the db entries, routes, and envs all need to use the exact same name
* Bumps the Supabase version to bring in the latest (significant) realtime refactor and pull in the newly functioning realtime migrations
* Creates `_realtime` schema on realtime pod start to segment realtime tables from the rest of the db
* Sets db `wal_level` to `logical` as supabase realtime uses the data produced by the WAL and requires a minimum `wal_level` of `logical` to get the data required for RLS to function


justinthelaw (approved):

justinthelaw (dismissed):

justinthelaw:
Recommend add tech debt for the migrations tracking!

barronstone (approved):
In Gato We Trust.

PR #863, opened 2024-08-02
fix: ensure whisper dev deps are installed when running make target

gphorvath:


gphorvath:
closing for now... just need to make sure these are installed before running the make targets.

PR #870, opened 2024-08-02
feat(api): openai compliant annotations with additional metadata

gphorvath:
* Adds vector_content ids to metadata returned by runs endpoints
* Adds an endpoint for querying vector_content to leapfrogai namespace - which contains chunk text


alekst23:
The annotations were not quite working with a simplistic test file that just said "Testing", so I had to augment it a little with some text data, and added a correlated question to the test file.

Perhaps there's better data we can use for this purpose.

alekst23:
I created a conformance test method that will evaluate the same feature in a separate call to OpenAI and LeapfrogAI:
`test_conf_chat_files.py > test_file_annotations`

This method tries to evaluate Annotations returned from an Assistant query, but will inherently also evaluate the features that build the Assistant in the process:
- Create VectorStore
- Create Thread
- Create and Poll Runs
- Create Assistant

We should set these up as separate tests on their own to isolate errors.

Presently, the OpenAI call succeeds and returns the proper type structure. 

LeapfrogAI call fails with a number of issues related to the structure of the message object in the _parameter_:
- id, missing
- content, must be list
- created_at, missing
- object, missing
- status, missing
- thread_id, missing

This fails on call to `client.beta.threads.create`

alekst23:
I have moved conformance testing for individual components to another PR and removed the code from here.

What's left should be a more minimal test focusing on conformance of file annotations.

Some issues that were identified via testing:
- threads.runs.create_and_poll - This should only attach messages to the run that were generated by the run itself, not any previous messages. 
- OpenAI annotations format is very simplistic and looks like this: 【4:0†source】
- LeapfrogAI annotations format is this: [file-uuid]

alekst23:
TODO: 
Reformat annotations:
`src/leapfrogai_api/backend/converters.py:57`

src/leapfrogai_api/backend/converters.py
```diff
@@ -44,24 +48,40 @@ def from_content_param_to_content(
...
+
+    if search_responses:
+        for search_response in search_responses.data:
+            all_file_ids += f"[{search_response.file_id}]"
+            all_vector_ids.append(search_response.id)
+            annotations.append(
                 FileCitationAnnotation(
-                    text=f"[{file_id}]",
-                    file_citation=FileCitation(file_id=file_id, quote=""),
+                    text="【4:0†source】",  # TODO: What should these numbers be? Who even knows...
```
justinthelaw, line 72: Maybe we should adapt this to our own purpose, even if it is not immediately clear. Let others correct us in open source community or when an OpenAI API-compatible library does end up using this differently or correctly.

My recommendation is:
```python
text=f"【{file_number}:{file_line}†{source}】"
```

Do you think this would still pass the conformance testing? If yes, then maybe we can stuff more demarcations into this format?

tests/conformance/test_chat_files.py
```diff
@@ -0,0 +1,106 @@
...
+        content="Who is Sam?",
+    )
+    run = make_test_run(client, assistant, thread)
+
+    messages = client.beta.threads.messages.list(
+        thread_id=thread.id, run_id=run.id
+    ).data
+
+    # Runs will only have the messages that were generated by the run, not previous messages
+    # assert len(messages) == 1
```
justinthelaw, line 94: Why is this commented out?

tests/conformance/test_chat_files.py
```diff
@@ -0,0 +1,106 @@
...
+
+def make_test_run(client, assistant, thread):
+    run = client.beta.threads.runs.create_and_poll(
+        assistant_id=assistant.id, thread_id=thread.id
+    )
+    return run
+
+
+def validate_annotation_format(annotation):
+    pattern = r"【\d+:\d+†source】"
```
justinthelaw, line 68: Is this always `†source`? I know we talked about this before, and I only remember that this entire field sometimes changes based on the stage at which OpenAI is at internally on the Chat message's RAG response (as observed in the network tab), prior it actually displaying to the UI.

justinthelaw (commented):
Some questions and comments for discussion!

src/leapfrogai_api/backend/converters.py
gphorvath: I don't think we actually have enough info today to capture file_number or file_line, but we could work on that.

tests/Makefile
```diff
@@ -74,3 +74,6 @@ test-api-unit:
 
 test-load:
 	python -m locust -f $$(pwd)/tests/load/loadtest.py --web-port 8089
+
+test-conformance:
+	source .env; PYTHONPATH=$$(pwd) pytest -vv -s tests/conformance
```
CollectiveUnicorn, line 79: Why does this need to source a `.env` file?

src/leapfrogai_api/backend/composer.py
```diff
@@ -138,22 +150,22 @@ def sort_by_created_at(msg: Message):
             vector_store_ids: list[str] = cast(list[str], file_search.vector_store_ids)
 
             for vector_store_id in vector_store_ids:
-                rag_responses: SearchResponse = await query_service.query_rag(
+                rag_responses = await query_service.query_rag(
```
CollectiveUnicorn, line 153: Why the typing removal?

CollectiveUnicorn (changes requested):

tests/Makefile
gphorvath: it doesn't but envs need to be set for API keys, URL, Model, etc. Probably should add some instructions on running conformance tests

src/leapfrogai_api/backend/composer.py
gphorvath: This is defined earlier now, so this would be redefining it

tests/conformance/test_chat_files.py
gphorvath: So this one took a little fiddling. I modified our implementation to return `filename` or `source` if for whatever reason `filename` isn't saved to the metadata in the vector_content table. OpenAI however doesn't have a predictable way to determine when they will spit out `filename` versus `source`, so I let the test pass if either is received.

tests/conformance/test_chat_files.py
gphorvath: Well... it turns out it's because we fail this test. That said, should I fix it as a part of this PR or add an issue?

justinthelaw: I think this can be fixed here since the test was written here too.

src/leapfrogai_api/backend/converters.py
gphorvath: I managed to squeeze in filename.

Unfortunately, it would require a pretty substantial refactor to get file_number (not actually sure what that means) or file_line (this is hard to extract in complex file types like PDF). That said, we could get to chunk_number fairly easily if we starting saving that info in the chunk metadata.

justinthelaw: Okay, maybe we can separate that into a different issue/PR then haha - I didn't mean to take you on a whirlwind, but I was just interested in how we might implement that in the future.

tests/Makefile
gphorvath: I added a little documentation on conformance tests to the README.md in the test directory.

tests/conformance/test_chat_files.py
gphorvath: ahhh, this actually lives in a different test, hence why it's commented here. but it fails there too

src/leapfrogai_api/backend/converters.py
gphorvath: Issue submitted: https://github.com/defenseunicorns/leapfrogai/issues/1110

CollectiveUnicorn (dismissed):

tests/Makefile
CollectiveUnicorn: This isn't a blocker to merging but do the conformance tests not use `load_dotenv` that would eliminate the need for explicitly sourcing wouldn't it? Also other tests require environment variables but AFAIK they just load them in within the Python code.

src/leapfrogai_api/data/crud_vector_content.py
```diff
@@ -65,6 +65,30 @@ async def add_vectors(self, object_: list[Vector]) -> list[Vector]:
         except Exception as e:
             raise e
 
+    async def get_vector(self, vector_id: str) -> Vector:
```
jalling97, line 68: I know this wouldn't fall under conformance testing, but should we have a test in place that validates the behavior of this vector endpoint? Perhaps something that just validates the fields we expect?

gphorvath: Sure thing! I'll add an e2e test

tests/Makefile
gphorvath: for sure. it looks like the unit tests still use load_dotenv, but they could easily just use environment variables

gphorvath:
Closing in favor of:

https://github.com/defenseunicorns/leapfrogai/pull/1165
https://github.com/defenseunicorns/leapfrogai/pull/1164

PR #883, opened 2024-08-06
chore(ui): typescript updates and fix rag e2e tests

### LINKED ISSUE #856: chore(ui): page server data type inference
In /routes/settings/api-keys/+page.svelte, the page gets data from the server (type PageServerLoad), but the page component is not automatically inferring it's type. Fix this and the red-squiggly warning should go away.

Check for other page data that has this issue and fix as well.

andrewrisse:
closes #856 
For the type inference issue, the data just needed to be named something else since there was a naming conflict.
Includes a few other updates to reduce typescript errors

There was a bad import in the Playwright rag.test.ts file causing those tests not to show up in the Playwright UI. This gave a false positive that all E2Es were passing. The rag tests were fixed to work correctly with Flowbite.

### DIFF
--- a/.github/workflows/e2e-llama-cpp-python.yaml
+++ b/.github/workflows/e2e-llama-cpp-python.yaml
@@ -4,9 +4,11 @@ name: e2e-llama-cpp-python
 on:
   pull_request:
     types:
-      - ready_for_review
-      - synchronize
-      - milestoned
+      - opened            # default trigger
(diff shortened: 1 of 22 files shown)


justinthelaw:
Looks like there's 1 UI lint error left on the pipeline - else everything looks okay to me

justinthelaw (approved):

PR #887, opened 2024-08-07
fix: batch requests to the CreateEmbedding stub

### LINKED ISSUE #885: Unable to attach certain PDFs to assistants
An assistant was created and two PDFs were attached to them (3MB and 13MB respectively). It appears that when the embeddings for the files are being created the the gRPC response containing all of the embedding vectors was to big resulted in an error in our API. After receiving the error the API would attempt to retry the request which would result in the same error occurring repeatedly.  This repeated retry-error loop would result in a situation where our embedding model would be pegged at 100% usage and our API would be slow to respond to additional requests.

YrrepNoj:
If we don't batch the requests we run the risk of erroring when the CreateEmbedding stub returns because the response is too large for our gRPC service to handle.

Closes #885

### DIFF
--- a/src/leapfrogai_api/backend/grpc_client.py
+++ b/src/leapfrogai_api/backend/grpc_client.py
@@ -1,6 +1,6 @@
 """gRPC client for OpenAI models."""
 
-from typing import Iterator, AsyncGenerator, Any
+from typing import Iterator, AsyncGenerator, Any, List
 import grpc
 from fastapi.responses import StreamingResponse
 import leapfrogai_sdk as lfai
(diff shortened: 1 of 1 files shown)


src/leapfrogai_api/backend/grpc_client.py
```diff
@@ -120,14 +120,25 @@ async def create_embeddings(model: Model, request: lfai.EmbeddingRequest):
     """Create embeddings using the specified model."""
     async with grpc.aio.insecure_channel(model.backend) as channel:
         stub = lfai.EmbeddingsServiceStub(channel)
-        e: lfai.EmbeddingResponse = await stub.CreateEmbedding(request)
+        embeddings: List[EmbeddingResponseData] = []
+
+        # Loop through inputs - 500 at a time
+        for i in range(0, len(request.inputs), 500):
```
justinthelaw, line 126: Is there rationale for the number 500 again? Is this close to the limit on the request object sent over the net?

CollectiveUnicorn (approved):

src/leapfrogai_api/backend/grpc_client.py
YrrepNoj: 500 was a somewhat arbitrary number I selected. This is something we can (and probably should?) look into adjusting in the future.


Background Context:
The original error we were getting was that the data in the response was a lot more than what the gRPC client was expecting. In my test case we had an input of 3852 request.inputs and the response was roughly 5x greater than what the gRPC client was expecting. I did some very naive math to come up with a batch size of 500 so that I knew we wouldn't run into this issue with this test case as well as allow a good amount of wiggle room for outlier inputs that might be even bigger.

justinthelaw: I think the thinking behind your chosen number is good enough as a conservative/functional choice IMO - optimization can come later! Just wanted us to record this somewhere for future reference. Thanks!

justinthelaw (approved):

Issue #590, opened 2024-06-04
feat(api,backend) load backend model info at runtime


CollectiveUnicorn:
### User Story

**As a** user of the API
**I want** to be able to load model metadata from the configs at runtime and access their information like type (embeddings/llm), dimensions, and precision  
**So that** I can use the appropriate models for my tasks

### Acceptance Criteria

**Given** a backend embeddings model configuration file  
**When** the API loads the model config
**Then** the API should be able to access the following information:
* Model type: "embeddings"
* Embedding dimensions
* Precision

**Given** a backend llm model configuration file  
**When** the API loads the model config
**Then** the API should be able to access the following information:
* Model type: "llm"

### Additional context

This feature will allow for more flexibility and dynamic model selection based on available resources and task requirements.

PR #890, opened 2024-08-07
feat(api, backend): Load and Expose Backend Model Info at Runtime

jamestexas:
## Summary
Fixes: #590

This PR introduces the ability to load and expose model metadata from configuration files at runtime within the API. This required somewhat significant refactoring of the `Config` class to allow easily. 
- The key changes before that refactoring can be found in this commit: ad3ddea3ae656cf1576b0b3c8c1b210cb55fdbee
- Refactoring and testing changes can be found here: 9727d5baed019e70f5d74711aea20ceab01884ea

I took care to retain function signatures with defaults if needed, and do not believe this should contain **any** breaking changes.

1. **Model Metadata Response**: Added a new `ModelMetadataResponse` class in `types.py` to encapsulate model metadata such as type, dimensions, precision, and capabilities.
2. **Config Enhancements**:
   - Introduced a `ModelMetadata` dataclass to represent metadata.
   - Updated `Config` class to include methods for loading and processing configuration files, including validation of metadata fields.
   - Ensured mutable default arguments are replaced with appropriate handling to avoid potential issues.
3. **API Route Update**: Modified the `models` endpoint in `models.py` to include metadata in the response.
4. **Testing Improvements**: Added tests for the new `Config` functionalities and validated correct handling of configuration files.
5. **Significant Refactoring for Testability**: Refactored the `Config` class to improve testability while retaining the original behavior.

#### Additional Context from internal Slack conversation:
- Confirmed metadata should be an object and discussed potential placement of `capabilities` within metadata.
- Highlighted the need for and added tests for the `Config` class, leveraging `pytest` for unit testing.

Testing coverage from this PR:
```shell
---------- coverage: platform darwin, python 3.12.4-final-0 ----------
Name                                 Stmts   Miss Branch BrPart  Cover   Missing
--------------------------------------------------------------------------------
src/leapfrogai_api/utils/config.py     130     13     35      0    90%   113-120, 140-141, 160-161, 228-229
--------------------------------------------------------------------------------
TOTAL                                  130     13     35      0    90%

```

## TODO:
- [ ] Update test dependencies to include `pytest` / `pytest-asyncio`. Is it appropriate to add dev dependencies into `/src/leapfrogai_api/pyproject.toml`?


src/leapfrogai_api/backend/types.py
```diff
@@ -54,6 +53,25 @@ class Usage(BaseModel):
...
+class ModelMetadataResponse(BaseModel):
+    type: Literal["embeddings", "llm"] | None = Field(
+        default=None,
+        description="The type of the model e.g. ('embeddings' or 'llm')",
+    )
+    dimensions: int | None = Field(
+        default=None,
+        description="Embedding dimensions (for embeddings models)",
+    )
+    precision: str | None = Field(
```
justinthelaw, line 65: We might want to constrain this further to only accept: float16, float32, bfloat16, int8, int4.

src/leapfrogai_api/backend/types.py
```diff
@@ -54,6 +53,25 @@ class Usage(BaseModel):
 ##########
 
 
+class ModelMetadataResponse(BaseModel):
+    type: Literal["embeddings", "llm"] | None = Field(
```
justinthelaw, line 57: Type and capabilities have cross-over based on the descriptions/examples here. We should be instead identifying modalities (image, text, speech) and capabilities (embeddings, chat, TTS, STT).

src/leapfrogai_api/backend/types.py
```diff
@@ -54,6 +53,25 @@ class Usage(BaseModel):
...
+        default=None,
+        description="The type of the model e.g. ('embeddings' or 'llm')",
+    )
+    dimensions: int | None = Field(
```
justinthelaw, line 61: This is based on pgvector limitations: https://github.com/pgvector/pgvector

This should use a validator (see Pydantic docs) to ensure it is no higher than 2000, even if the model supports more.

justinthelaw (commented):
Looks great so far! I'll do a deeper dive once it's fully ready. Just some comments to propagate downwards and discuss. Thanks!

src/leapfrogai_api/backend/types.py
jamestexas: Hmm, should we maybe differentiate on this too? Would it make sense to alias `type` to `modalities` (to encapsulate what you said) and then adding an enum for capabilities?
I don't know how `type` is used outside of this context and want to make sure it conforms to the rest of the repo.

src/leapfrogai_api/backend/types.py
justinthelaw: Also brings up a missing field: format. This includes None, GPTQ, GGUF, SqueezeLLM, AWQ. This depends on whether the model was quantized, and whether it is compatible with vLLM and/or llama.cpp

jamestexas:
> Looks great so far! I'll do a deeper dive once it's fully ready. Just some comments to propagate downwards and discuss. Thanks!

Outside of the PR feedback is there anything else missing? This worked in terms of interacting with the API and wasn't sure if I needed to update things anywhere else to make it "complete".  Let me know!

src/leapfrogai_api/backend/types.py
justinthelaw: I guess I was thinking there are vision embeddings-only models versus vision comprehension-only (chat), for example. So the combination of modality and capability works; however, if you're saying type is basically modality, then I am okay with that. Modality is more of an industry-standard term (e.g., how HF organizes models using semantic search: https://huggingface.co/models)

justinthelaw:
> > Looks great so far! I'll do a deeper dive once it's fully ready. Just some comments to propagate downwards and discuss. Thanks!
> 
> Outside of the PR feedback is there anything else missing? This worked in terms of interacting with the API and wasn't sure if I needed to update things anywhere else to make it "complete". Let me know!

This PR is perfectly constrained to the issue's main concerns, as long as the tests are there for the new config and fields. Areas like UI consuming config and Backends producing config are meant for future tickets.

src/leapfrogai_api/backend/types.py
jamestexas: @justinthelaw I think I am a bit unsure, to be honest! Referencing [this config map](https://github.com/defenseunicorns/leapfrogai/blob/9eba6791f1b0852cf03474d7a0fd9339a90deae7/packages/vllm/chart/templates/configmap.yaml#L11) I "lifted" type out of what was defined already in the config, based off [of this convo](https://defense-unicorns.slack.com/archives/C06H8EVTG0M/p1723047375497539).  Tentatively looking at an implementation like this though (and will proceed with this for now unless you have any opinions otherwise):

```python
class ModelMetadataResponse(BaseModel):
    """Metadata for the model, including type, dimensions (for embeddings), and precision."""
    model_config = ConfigDict(use_enum_values=True)

    capabilities: list[Capability] | None = Field(
        default=None,
        description="Model capabilities (e.g., 'embeddings', 'chat', 'tts', 'stt')",
    )
    dimensions: int | None = Field(
        default=None,
        description="Embedding dimensions (for embeddings models)",
    )
    format: Format | None = Field(
        default=None,
        description="Model format (e.g., None, 'GPTQ', 'GGUF', 'SqueezeLLM', 'AWQ')",
    )
    modalities: list[Modality] | None = Field(
        default=None,
        description="The modalities of the model (e.g., 'image', 'text', 'speech')",
    )

    precision: Precision | None = Field(
        default=None,
        description="Model precision (e.g., 'float16', 'float32', 'bfloat16', 'int8', 'int4')",
    )
    type: Literal["embeddings", "llm"] | None = Field(
        default=None,
        description="The type of the model e.g. ('embeddings' or 'llm')",
    )
```

src/leapfrogai_api/routers/openai/models.py
```diff
@@ -18,8 +24,14 @@ async def models(
 ) -> ModelResponse:
     """List all available models."""
     res = ModelResponse(data=[])
-    model_config: Config = get_model_config()
-    for model in model_config.models:
-        m = ModelResponseModel(id=model)
+    # shared config object from the app
+    model_config: "Config" = session.app.state.config
```
CollectiveUnicorn, line 28: Seems like there might be an issue here, the `supabase_session` object doesn't have a property named app so its failing for me when I hit this endpoint: https://leapfrogai-api.uds.dev/openai/v1/models

Screenshot 2024-08-28 at 3 20 59 PM

jamestexas:
@CodiumAI-Agent /describe
--pr_description.publish_description_as_comment=true

gphorvath:
As we icebox the repo, we are leaving this draft PR as a good place to pickup if we resume work. The goal behind the work was to expose model information such as type (embeddings, chat, etc.) so that the API can infer the model backend without requiring additional state.

PR #896, opened 2024-08-09
chore: Use template function to set model PVC names

YrrepNoj:
This PR updates our PVCs and Deployments to use a helper function to determine the name of the PVC based on the model and revision used in the Zarf Package.

This update buys us the ability to deploy multiple models to the cluster without clashing on the name of the PVC they were trying to use.

- Adds `ZARF_CONFIG` environment variable to our Make targets for when we build model Zarf packages.
  - This is necessary because the `actions.before` lifecycle hook needs certain templates applied so it can download the correct model weights at package-create-time.
  - This gives package creators the ability to choose a different model to bake into their package.


packages/vllm/zarf.yaml
```diff
@@ -53,7 +57,5 @@ components:
           # NOTE: This assumes python is installed and in $PATH and 'huggingface_hub[cli,hf_transfer]' has been installed
           - cmd: python src/model_download.py
             env:
-              - LAI_REPO_ID=TheBloke/Synthia-7B-v2.0-GPTQ
-              - LAI_REVISION=gptq-4bit-32g-actorder_True
-              - LAI_QUANTIZATION=gptq
-              - LAI_HF_HUB_ENABLE_HF_TRANSFER=1
```
justinthelaw, line 59: `LAI_HF_HUB_ENABLE_HF_TRANSFER` and `LAI_QUANTIZATION` need to be transferred over to the Zarf config and the Zarf package manifest, as AppConfig (Confz) requires it.

In the future, quantization can be removed (vLLM automatically detects and optimizes from the model files' config.yaml) and hf_transfer can just be hardcoded to always be on.

justinthelaw (changes requested):

packages/vllm/zarf.yaml
YrrepNoj: Thanks for reminding me, I added this back.

YrrepNoj:
This is stale. Will reopen once I pick this up again.

PR #904, opened 2024-08-13
Update README.md

thryothorus:


justinthelaw:
The API no longer recommends using the supabase CLI in upcoming docs overhaul.

Issue #911, opened 2024-08-14
bug(api): Vectors are not properly being chunked and stored in the db


CollectiveUnicorn:
### Steps to reproduce
1. Upload a large file with 1000+ words
2. Create a vector store
3. Index the vector store

### Expected result
Chunks in `VectorStoreContent` are added for all of the data.

### Actual Result
A small subset of chunks are added to the db in `VectorStoreContent`

### Visual Proof (screenshots, videos, text, etc)
This should contain 40+ entries for a single file but only shows a fraction of that for multiple files:
![Screenshot 2024-08-14 at 2 53 45 PM](https://github.com/user-attachments/assets/f9266359-fc45-4b54-af65-ab5b49a557bb)

### Severity/Priority
High

### Additional Context
This prevents RAG from functioning properly

CollectiveUnicorn:
Resolved by https://github.com/defenseunicorns/leapfrogai/pull/912

PR #912, opened 2024-08-14
bug(api): Fix issue where only every 100th vector was being stored #911

vanakema:
Change was tested locally and is verified to fix the issue #911


justinthelaw (approved):
Sharp eyes!

PR #917, opened 2024-08-15
chore(api): Generate API keys to use for backend API calls

CollectiveUnicorn:
* Updates the API key logic so that whenever a request is made to the backend API with a JWT, the auth layer will create an API key to use to interact with Supabase. This is to address issues with tokens expiring during long running processes or as a result of invalidation on the frontend. These temporary (1 hour lifespan) keys will ensure that processes are authed for a sufficient amount of time.
* Adds trigger to delete expired API keys on insert or update.
* Refactors supabase_session auth code and updates related tests.


CollectiveUnicorn:
No longer needed

PR #933, opened 2024-08-19
chore!: standardize flavors, bundle dir and ghcr namespace

justinthelaw:
## OVERVIEW

Moves bundles to UDS Common standard directory and shifts the target GHCR repository sub-directory from `packages/leapfrogai` to `packages/uds/leapfrogai` as part of the [_Made For UDS_](https://github.com/defenseunicorns/uds-common/blob/main/docs/uds-package-practices.md) standard

Moving forward, our packages will be published at a different sub-directory, so at the next release, all bundles must be updated to reflect these.

Additionally, as packages will be by-default published with a flavor of `upstream` the `latest` bundles must point to the correct flavor as published by `uds zarf package publish`. Zarf publishing will append `-${FLAVOR}` to the package when it is uploaded to the registry, and the bundle must use the correct ref. Due to this, after 0.12.0 is released and artifacts are published, PR #977 should be merged in to add the UDS bundle manifest pointers necessary to use the `upstream` flavor and v0.12.0 in general.

## BREAKING CHANGES

- LeapfrogAI `latest` UDS bundles on `main` branch will break when this is merged unless the following is done:
  - add `-upstream` to all components in `latest` bundles
  - change namespace of the referenced repository to `packages/uds/leapfrogai`
- release-please does not currently work with flavors as it is currently configured to follow SemVer standards; manual changing of the `latest` bundles ref must be done moving forward until a fix (custom replacement strategy) is implemented

## CHANGES

- Moves bundles to UDS Common standard directory (`bundles/`)
- Shifts the target GHCR repository sub-directory from `packages/leapfrogai` to `packages/uds/leapfrogai` as part of the[ _Made For UDS_](https://github.com/defenseunicorns/uds-common/blob/main/docs/uds-package-practices.md) standard
- Adds `-upstream` as the default bundle ref and increments the ref manually


justinthelaw:
Relies on #920

bundles/latest/cpu/uds-bundle.yaml
```diff
@@ -0,0 +1,38 @@
...
+metadata:
+  name: leapfrogai
+  description: A UDS bundle for deploying LeapfrogAI
+  version: 0.12.0-upstream
+
+packages:
+  # Supabase backend for the UI and API to interface with Postgresql
+  - name: supabase
+    repository: ghcr.io/defenseunicorns/packages/uds/leapfrogai/supabase
+    ref: 0.12.0-upstream
```
YrrepNoj, line 13: I see what we're doing here (using the ref 0.12.0 that isn't out yet because this is where it _will soon_ be published to, but what are your thoughts on using the same config bundle for 0.11.0 so people checking out main can still build the latest bundle and then creating a new PR for this configuration before we do our 0.12.0 release?

YrrepNoj (dismissed):
Looks good to me. I did open a pondering question to see if we might want to keep our main-latest-bundles buildable.

I think I am leaning towards keeping main buildable and opening a new PR with the change in namespace when we get closer to releasing and I'm curious what others might think.

bundles/latest/cpu/uds-bundle.yaml
justinthelaw: Ah okay! So just to clarify: would you like me to rollback the `0.12.0-upstream` and set it back to 0.11.0, and put the `0.12.0-upstream` change on a different branch for post 0.12.0-release?

YrrepNoj: I think that would be a good way to handle this type of update 👍

justinthelaw: Okay! Also, I didn't mention, but I will also rollback the bundle's pointer to `packages/uds/leapfrogai` as well, and hold it until post-0.12.0 release.

YrrepNoj (approved):

CollectiveUnicorn (approved):

Issue #706, opened 2024-07-01
chore(lint): uds and zarf lints


justinthelaw:
### Describe what should be investigated or refactored

UDS (bundle and task) and Zarf linting should be pinned and applied to the pre-commit and GitHub CI workflows.

- [ ] UDS bundle lint
- [ ] UDS task lint
- [ ] Zarf package lint
- [ ] (OPTIONALLY) Helm chart directory lint
- [ ] (OPTIONALLY) Docker manifest lint

### Links to any relevant code

Example pre-commit: https://github.com/justinthelaw/uds-rke2/blob/main/.pre-commit-config.yaml

Example linting workflows:

- https://github.com/justinthelaw/uds-rke2/blob/main/.github/workflows/zarf-lint.yaml
- https://github.com/justinthelaw/uds-rke2/blob/main/.github/workflows/uds-lint.yaml
- (OPTIONALLY) https://github.com/helm/chart-testing-action/blob/main/action.yml (Helm)
- (OPTIONALLY) https://github.com/hadolint/hadolint (Docker)

### Additional context

N/A

Issue #705, opened 2024-07-01
chore(helm): audit zarf variable usage


justinthelaw:
### Describe what should be investigated or refactored

Ensure as few Zarf variables as possible, instead opting for deployment via UDS overrides on Helm charts and values files.

### Links to any relevant code

`Made for UDS` applicable `SHOULD` requirement: https://github.com/defenseunicorns/uds-common/blob/main/docs/uds-package-practices.md#structure

### Additional context

N/A

Issue #367, opened 2024-04-08
chore: Slim down the things getting installed into the the vLLM Dockerfile


YrrepNoj:
### Describe what should be investigated or refactored
Right now we are installing several things into our vLLM container that we are not installing into the other model containers. Most of these installations are necessary, but some could potentially be removed.

Go through the things being installed (especially things being installed via apt-get) and either:
1) Remove the installation if it is unnecessary
2) Leave a comment describing WHY it is necessary to install 

### Links to any relevant code
- https://github.com/defenseunicorns/leapfrogai/blob/main/packages/vllm/Dockerfile

PR #937, opened 2024-08-20
chore!: add containerization and packaging manifest lints

### LINKED ISSUE #365: chore: pre-commit updates and improvements
We need to add some more pre-commits to align with more best practices, including Dockerfile, Zarf, UDS, and Helm chart linting.

https://github.com/justinthelaw/repository-template

### Additional context
N/A

### LINKED ISSUE #965: bug: vllm bundle has invalid kubernetes spec and won't properly deploy on internal infrastructure
### Environment

1. OS and Architecture: RHEL8 Kubernetes on rke2
2. App or Package Name: vllm
2. App or Package Version: 0.11.0
3. Kubernetes Distribution: rke2
4. Kubernetes Version: rke2 1.28.12
6. Other: This bug is showing up in our internal Defense Unicorns Infrastructure

### Steps to reproduce

1. Deploy vllm bundle 0.11.0

### Expected result
vllm zarf package should deploy successfully
-

### Actual Result
zarf deployment hangs and is unable to create the data injection target
-

### Visual Proof (screenshots, videos, text, etc)
image
```
W0829 00:07:19.536249     407 warnings.go:70] unknown field "spec.template.spec.containers[0].securityContext.fsGroup"
W0829 00:07:19.536261     407 warnings.go:70] unknown field "spec.template.spec.initContainers[0].securityContext.fsGroup"
  •  Processing helm chart vllm-model:0.11.0 from Zarf-generated helm chart
 NOTE  Using config file
mkdir: can't create directory '/data/.model': Permission denied
command terminated with exit code 1
 WARNING  Unable to create the data injection target directory /data/.model in pod
 vllm-model-5c99c4fc68-xpql5
 NOTE  Using config file
mkdir: can't create directory '/data/.model': Permission denied
command terminated with exit code 1
 WARNING  Unable to create the data injection target directory /data/.model in pod
 vllm-model-5c99c4fc68-xpql5
```

### Additional Context

There is actually an invalid kubernetes manifest spec that may be causing this.  Defining fsGroup under initContainers or containers is not valid, and therefore that may be the issue creating this directory.  Instead, fsGroup should be defined at the spec.template.spec.securityContext level, not within the containers/initContainers array.
Improper definitions can be found [here](https://github.com/defenseunicorns/leapfrogai/blob/v0.11.0/packages/vllm/chart/values.yaml#L26), [here](https://github.com/defenseunicorns/leapfrogai/blob/v0.11.0/packages/vllm/chart/templates/deployment.yaml#L37) and [here](https://github.com/defenseunicorns/leapfrogai/blob/v0.11.0/packages/vllm/chart/templates/deployment.yaml#L64).

justinthelaw:
## Description

Adds linting for Helm, Docker, Zarf, and UDS manifests to improve the quality and consistency of all containerization and Kubernetes related logic.

Adds workflows and pre-commits to ensure consistency before and after committing to a PR.

Remaining lower-priority Dockerfile lint errors are in issue #984 

### BREAKING CHANGES

- `GPU_CLASS_NAME` is now `GPU_RUNTIME` to better align with the actual manifest spec field and purpose
  - **RATIONALE:** https://github.com/defenseunicorns/leapfrogai/pull/937#discussion_r1745914222

### CHANGES

- Adds Dockerfile linitng
- Adds Helm manifest linting
- Adds UDS bundle linting
- Adds Zarf package linting
  - Moves Zarf variables to override values files (`upstream-values.yaml`)
- Centralizes linting schemas and configs at the root of the project
  - UDS and Zarf lints rely on a downloaded schema from upstream repository [uds-cli](https://github.com/defenseunicorns/uds-cli)
  - Pre-commit checks for and locally downloads for the schema, schema is gitignored
  - Workflow downloads the schema file fresh every time
  - Schemas are pinned to our UDS CLI version of `v0.14.0`
  - Dockerfile linting config is defined in the .hadolint.yaml
  - Helm linting uses the [helm binary lint](https://helm.sh/docs/helm/helm_lint/)
- Refactors applicable manifests based on linting errors and warnings
  - "Warning" and "Info" level issues are considered "passing", but can be viewed as outputs in workflow logs (issue #984)
  - All component Chart packages have `values.yaml` with default values, for UDS overrides following UDS Core patterns
  - A portion of the Dockerfile changes from #835 are integrated here to remove linting warnings, and reduce unnecessary layers and dependencies
  - Moves `env` variables to values files
  - Moves basic pod level configurations to values files
  - Standardizes existing/basic values across all charts
  - Removes autoscaling values that are not tied to a [metrics-server](https://github.com/kubernetes-sigs/metrics-server#readme) implementation nor [HorizontalPodScaling](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#creating-the-autoscaler-declaratively) resource definition
- Removes `GPU_ENABLED` python code/env from llama-cpp-python
  - CPU-only package and image (no CUDA deps)
  - No exposure or setting of `GPU_ENABLED` via values nor Zarf variables
  - Was previously always set to `0` anyway, with no kosher way to change it
- Adds docs related to UDS bundle overrides based on the new values files
- Deactivates `e2e_registry1_weekly` test on draft PRs (mostly to stop it from running here all the time)
- Removes duplicate `check-merge-conflict` pre-commit
- More comprehensive `clean` Make target for build artifact cleaning
- Updates the Registry1 weekly to `0.12.2`

## Related Issue

Fixes #965 #706 #705 #367 #365

## Checklist before merging


justinthelaw:
Instead of using an [outdated and/or sketchy helm-lint action](https://github.com/justinthelaw/repository-template/blob/main/.github/workflows/helm-lint.yaml) use [this](https://github.com/Azure/setup-helm) to manually install helm v3.15.1 and manually perform a `helm lint [sub-directory]`.

justinthelaw:
Remaining components to fix:
1. vLLM Dockerfile missing variable
2. llama-cpp-python injection failing
3. UI helm chart reorganization
4. Supabase chart reorganization

bundles/latest/gpu/uds-config.yaml
```diff
@@ -1,11 +1,11 @@
 # see individual zarf packaging configuration for more variables and variable descriptions
 variables:
   text-embeddings:
-    gpu_class_name: "nvidia"  # Set to ensure the nvidia runtimeClass is present in case GPU limit is increased
+    GPU_RUNTIME_NAME: "nvidia"  # Set to ensure the nvidia runtimeClass is present in case GPU limit is increased
```
justinthelaw, line 4: Change to lowercase within the UDS bundle configS.

Change to `GPU_RUNTIMECLASSNAME`.

justinthelaw (commented):
Fix UI UDS package CR for KeyCloak SSO, and test locally again.

packages/whisper/chart/templates/deployment.yaml
```diff
@@ -2,16 +2,14 @@ apiVersion: apps/v1
...
 spec:
-  strategy:
-    rollingUpdate:
-      maxUnavailable: 0
-    type: RollingUpdate
-  {{- if not .Values.autoscaling.enabled }}
   replicas: {{ .Values.replicaCount }}
+  {{- with .Values.strategy }}
+  strategy:
+    {{- toYaml . | nindent 8 }}
```
YrrepNoj, line 12: Similar comment here as elsewhere.

Is there a particular reason to use the `{{- with` clause instead of `{{- toYaml listing.the.path.here`?

packages/ui/values/upstream-values.yaml
```diff
@@ -0,0 +1,47 @@
+image:
+  repository: ghcr.io/defenseunicorns/leapfrogai/leapfrogai-ui
+  tag: ###ZARF_CONST_IMAGE_VERSION###
```
YrrepNoj, line 3: I'm not going to comment on all of them but I think `###ZARF_VAR/CONST` should be within quotes. Especially in places like this.

I have seen issues in the past where the tag is not able to be templated correctly because the tag is a number when trying to concat with the repository string later in the deployment.

YrrepNoj (changes requested):
Looked through most of the changes and heavily agree with the general direction this PR is going.

Have a few questions and tweaks to request though.

packages/whisper/zarf.yaml
```diff
@@ -16,8 +16,8 @@ variables:
     description: The GPU limit for the model inferencing.
     default: "0"
     pattern: "^[0-9]+$"
-  - name: GPU_CLASS_NAME
-    description: The GPU class name for the model inferencing. Leave blank for CPU-only.
+  - name: GPU_RUNTIME
```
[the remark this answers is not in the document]
justinthelaw: I originally wanted to completely match the actual field, `GPU_RUNTIME_CLASS_NAME` but decided that it was too long, so I shortened it to `GPU_RUNTIME`. `CLASS_NAME` is much more vague than `RUNTIME`, since classname could be interpreted as the GPU class (Ada, etc.), whereas runtime is the proper word to define the actual underlying container software (`nvidia`). For example, if you run `docker info` or look at `containerd` config tomls, `runc` and `nvidia` are called `runtimes` and not classes.

packages/whisper/values/upstream-values.yaml
```diff
@@ -1,8 +1,9 @@
 image:
-  tag: "###ZARF_CONST_IMAGE_VERSION###"
+  repository: ghcr.io/defenseunicorns/leapfrogai/whisper
+  tag: ###ZARF_CONST_IMAGE_VERSION###
 
 gpu:
-  runtimeClassName: "###ZARF_VAR_GPU_CLASS_NAME###"
+  runtimeClassName: ###ZARF_VAR_GPU_RUNTIME###
```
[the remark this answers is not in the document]
justinthelaw: This was an unintentional removal of the quotes. You are right that these are supposed to be a string, and adding explicit quotes would cast it to a string for sure. Will change!

packages/whisper/chart/values.yaml
```diff
@@ -48,13 +52,38 @@ autoscaling:
...
+    maxUnavailable: 0
+  type: RollingUpdate
+
 persistence:
   size: 15Gi
   accessModes: ReadWriteOnce
   storageClass: "local-path"
+
+gpu:
+  runtimeClassName: "###ZARF_VAR_GPU_RUNTIME###"
```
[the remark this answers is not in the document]
justinthelaw: You are right! I missed these - my bad

.github/workflows/zarf-lint.yaml
```diff
@@ -0,0 +1,83 @@
+name: Zarf Lint
```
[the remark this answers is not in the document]
justinthelaw: I wanted to have the right triggers in each workflow, but I guess we could have `on` triggers per-job - I don't think I have done that before?

.github/workflows/uds-lint.yaml
```diff
@@ -0,0 +1,54 @@
+name: UDS Lint
+
+on:
+  push:
+    branches:
+      - "main"
```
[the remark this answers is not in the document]
justinthelaw: That is true! I think that is an artifact of my copy/paste from my repository-template, where I was experimenting.

packages/api/Dockerfile
```diff
@@ -21,9 +21,11 @@ RUN rm -f packages/api/build/*.whl && \
     pip install packages/api/build/leapfrogai_api*.whl --no-index --find-links=packages/api/build/
 
 FROM ghcr.io/defenseunicorns/leapfrogai/python:3.11
-ENV PATH="/leapfrogai/.venv/bin:$PATH"
+
 WORKDIR /leapfrogai
 
+ENV PATH="/leapfrogai/.venv/bin:$PATH"
```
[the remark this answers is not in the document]
justinthelaw: That was actually a mistake when I was experimenting! I can revert to reduce this PR's changes

packages/api/chart/templates/migration-job.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: batch/v1
 kind: Job
 metadata:
   name: {{ include "chart.fullname" . }}-migrations-{{ .Values.api.migration.image.tag | default .Chart.AppVersion }}
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ default "leapfrogai" .Release.Namespace }}
```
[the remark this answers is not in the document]
justinthelaw: 1. No, helm installs where `--name` isn't provided will provide a Helm-generated name, so technically it would never default to `leapfrogai`
2. That being said, I still like the idea of providing the default as an indicator to a user that `leapfrogai` is the default (e.g., the UI will not work out-of-the-box if the namespace changes due to the default `LEAPFROGAI_API_BASE_URL`. I will change it back to our original pattern, since it was mostly a preference thing on my part. The functionality of both syntaxes is the same!

packages/vllm/chart/templates/deployment.yaml
```diff
@@ -63,6 +54,10 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- with .Values.env }}
+          env:
+            {{- toYaml . | nindent 12 }}
```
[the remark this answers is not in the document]
justinthelaw: The way the values are rendered when using the method you outlined fails to render when using Zarf. I can't explain why, because `helm template --dry-run` works as-is.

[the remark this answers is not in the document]
justinthelaw: I have seen the `with` method used in other OSS charts, and it isn't taboo or wrong in any sense, just a diff pattern

packages/whisper/chart/templates/deployment.yaml
justinthelaw: See https://github.com/defenseunicorns/leapfrogai/pull/937#discussion_r1745933590

packages/ui/values/upstream-values.yaml
justinthelaw: I can do that! This was mostly an inconsistency and my preference - my bad

gscallon:
Like the UDS overrides pattern with the default values for the component charts 👍

gscallon (dismissed):

justinthelaw:
@YrrepNoj @gscallon some of the changes made for comment resolution, and also the merge with main, broke the tests. Will fix everything early tomorrow morning and ask for re-review.

justinthelaw:
> Awaiting approval of 0.12.2 on IronBank. Resolving several new CVEs in VAT.
> 
> https://repo1.dso.mil/dsop/opensource/defenseunicorns/leapfrogai/api/-/merge_requests/32
> 
> https://vat.dso.mil/vat/image?imageName=opensource/defenseunicorns/leapfrogai/api&tag=v0.12.2&branch=renovate/defenseunicorns-leapfrogai-0.x

Same as #977

.github/workflows/zarf-lint.yaml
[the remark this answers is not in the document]
jalling97: Personally I think it's easier to track with all of the linting types being separate, so unless we have an upper limit to the number of workflows we should have (or something else I'm missing), I don't think it makes a huge difference so I think it works as is.

.github/workflows/zarf-lint.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Download Zarf Package Schema
+        # TODO: renovate setup
+        run: curl -o zarf.schema.json https://raw.githubusercontent.com/defenseunicorns/uds-cli/v0.14.0/zarf.schema.json
+
+      - name: Install jsonschema
+        run: pip install check-jsonschema==0.28.0
+
+      - name: Validate API zarf.yaml
+        if: always()
```
jalling97, line 40: For my own education, are the `if: always()` necessary? I would assume that excluding them would result in the each `run` command always running, right?

justinthelaw: I did that on purpose so that all of them run and show all possible errors in minutes edited by the developer. Unless you think it's better to not do that?

justinthelaw: Failures will still fail the check at any of those steps in the job. I just thought it would be better user experience to show all errors in all manifests instead of stopping the job short when there is one failure.

jalling97: Ah I see now. No I think that makes sense, I just didn't understand the purpose at first. Thanks for clarifying!

.pre-commit-config.yaml
```diff
@@ -76,3 +80,60 @@ repos:
...
+  # UDS v0.14.0
+  #############
+
+  # ZARF PACKAGE CHECK
+  - repo: local
+    hooks:
+      - id: download-schema
+        name: "Download Zarf Schema"
+        entry: |
+          bash -c 'FILE="zarf-v0.14.0.schema.json"
```
jalling97, line 101: Hmm is there anyway to easily pin this version somewhere so we don't have to track down each `v0.14.10` when we need to update UDS versions? I haven't used YAML anchors and aliases before but I wonder if we could use an anchor to set the UDS version and alias it throughout? Thoughts?

justinthelaw: Ah yes, I thought about pinning all of our development deps in a YAML and then using `uds zarf tools yq` to read them (e.g., UDS CLI, k3d, PyEnv etc.), but I think that can be a refactor for future dev-experience optimizations.

justinthelaw: Another possibility is the use of `renovate` (like in UDS Core, or our IronBank repositories), but that is a much larger task and refactor.

packages/text-embeddings/Dockerfile
```diff
@@ -12,14 +12,14 @@ RUN python3.11 -m venv .venv
 ENV PATH="/leapfrogai/.venv/bin:$PATH"
 
 # copy and install all python dependencies
-# NOTE: We are copying the leapfrogai whl to this filename because installing 'optional extras' from
+# NOTE: We are copying the leapfrog whl to this filename because installing 'optional extras' from
```
jalling97, line 15: Super pedantic lol, why the change from `leapfrogai` to `leapfrog`?

jalling97 (dismissed):
Awesome job cleaning things up!

packages/text-embeddings/Dockerfile
justinthelaw: Oh good catch! I can change that back. I think this was an artifact of changes I made in a previous PR not being reflected in this forked-branch where I brought all these changes in from.

jalling97 (dismissed):

vanakema:
some awesome cleanup, ty justin. I didn't review the code closely, but thanks for giving the thorough changelog. This is all great stuff in making the "ux" of contributing and deploying lfai better

justinthelaw:
Merge this one in first: https://github.com/defenseunicorns/leapfrogai/pull/977

Then change e2e registry1 weekly to point to the commit SHA as a result of the above merge.

CollectiveUnicorn (approved):

PR #939, opened 2024-08-21
fix(api): search vectorstore using only last message

alekst23:
## Description

Adjust the RAG implementation to search the vector store using the last user message instead of the first.

Add a needle in a haystack test to confirm problem and fix.

Apply fix for indexing already existing files to return the found file instead of 500.


tests/integration/api/test_rag_files.py
```diff
@@ -0,0 +1,101 @@
...
+        content="There is one piece of fruit in the fridge. What is it and where is it located?",
+    )
+    run = make_test_run(client, assistant, thread)
+
+    messages = client.beta.threads.messages.list(
+        thread_id=thread.id, run_id=run.id
+    ).data
+
+    # Get the response content from the last message
+    message_content = messages[-1].content[0].text
```
jalling97, line 96: I don't think we can assume that the response is always going to be the last item in the list of messages. I found that it can occasionally get swapped with the user message, so I've had to search through the messages to find the actual response.

alekst23:
There's one issue to call out here, which is the changes to `src/leapfrogai_api/backend/rag/index.py`. I was encountering an error during testing where the file to be attached to an assistant was already in the vector store. In this case, the endpoint was returning error 500. I don't have an OpenAI conformance test for this yet, but it did not have the same problem during testing.

tests/integration/api/test_rag_files.py
alekst23: This would be a bug with our API that we would need to resolve.

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -53,12 +53,12 @@ async def index_file(self, vector_store_id: str, file_id: str) -> VectorStoreFil
...
 
-        if await crud_vector_store_file.get(
+        if file_existing := await crud_vector_store_file.get(
             filters=FilterVectorStoreFile(vector_store_id=vector_store_id, id=file_id)
         ):
-            logging.error("File already indexed: %s", file_id)
-            raise FileAlreadyIndexedError("File already indexed")
-
+            # logging.error("File already indexed: %s", file_id)
+            # raise FileAlreadyIndexedError("File already indexed")
```
CollectiveUnicorn, line 60: Can we get rid of these comments?

CollectiveUnicorn (changes requested):

tests/integration/api/test_rag_files.py
```diff
@@ -0,0 +1,101 @@
...
+from openai.types.beta.threads.text import Text
+from openai import OpenAI
+
+
+LEAPFROGAI_MODEL = "llama-cpp-python"
+OPENAI_MODEL = "gpt-4o-mini"
+
+
+def text_file_path():
+    return Path(os.path.dirname(__file__) + "/../../data/test_with_data.txt")
```
YrrepNoj, line 12: This function isn't being used and this file isn't a part of this PR. Was this a copy-paste from somewhere else?

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -53,12 +53,12 @@ async def index_file(self, vector_store_id: str, file_id: str) -> VectorStoreFil
         crud_vector_store_file = CRUDVectorStoreFile(db=self.db)
         crud_vector_store = CRUDVectorStore(db=self.db)
 
-        if await crud_vector_store_file.get(
+        if file_existing := await crud_vector_store_file.get(
```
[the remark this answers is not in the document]
YrrepNoj: I realize this is beyond the scope of the bug fix you are implementing here, I wanted to leave a comment here so I didn't forget, hopefully spark a little conversation to check my assumptions, and then make a new issue for our backlog.

tests/integration/api/test_rag_files.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def leapfrogai_client():
+    return OpenAI(
+        base_url=os.getenv("LEAPFROGAI_API_URL"),
+        api_key=os.getenv("LEAPFROGAI_API_KEY"),
+    )
+
+
+def client_config_factory(client_name):
```
YrrepNoj, line 26: This factory makes it hard to type-hint the returned object (as well as type hint the values that the different keys refer to when they're used later). 

What are your thoughts on a wrapper class our typed-tuple return instead of a dict?

YrrepNoj (changes requested):

src/leapfrogai_api/backend/rag/index.py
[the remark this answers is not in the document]
alekst23: Not sure if this is the right place for that type of validation, I think it may need to be done upstream. The reason being is that this function simply tries to add the file to the vector store if it does not exist. It can fail for various reasons, and the file will still be in the database with a status like "canceled" or "failed", and returned to caller. If we try to add it again, it will be duplicated without the user/caller being aware.

I think if there are issues with the file status, it needs to be deleted and re-sent (maybe), but that's up to the calling function.

tests/integration/api/test_rag_files.py
alekst23: True. This was part of a simpler test version. Removed.

tests/integration/api/test_rag_files.py
alekst23: That's a good idea.

This factory is actually part of conformance testing PR that I re-used here, so I have consolidated the code in a `tests/utils/client.py` module and refactored all the relevant test files.

justinthelaw (approved):

CollectiveUnicorn (approved):

CollectiveUnicorn (approved):

YrrepNoj (approved):

Issue #789, opened 2024-07-16
feat(api): Chat with existing files while indexing new files


CollectiveUnicorn:
### User Story

**As the** LeapfrogAI UI
**I want** to chat with existing documents while new documents are indexing
**So that** the user does not have to wait for the new docs to be added before being able to continue their work

### Acceptance Criteria

**Given** there is a vector store with existing documents tied to an assistant
**When** new documents are added and indexed
**Then** the user can still chat with their existing document via the `runs` endpoint without waiting for the indexing of the new files to finish

andrewrisse:
once #885 is merged, we can test to see if this is still an issue

CollectiveUnicorn:
> once #885 is merged, we can test to see if this is still an issue

Unfortunately, it seems like this is still an issue.

Issue #626, opened 2024-06-13
feat(backend) Implement Queue for computationally expensive tasks


barronstone:
### User Story

**As a** LeapfrogAI end user
**I want** to upload multiple documents at once
**So that** so that I can create Assistants to "chat with my docs"

**As a** LeapfrogAI end user
**I want** to upload multiple audio and/or video files at once
**So that** so that I can transcribe and/or translate the recording

### Acceptance Criteria

**Given** the user has multiple files to upload for RAG and/or transcription/translation
**When** the user uploads those files via the UI or API
**Then** the files are queued up for sequential in-processing

**Given** multiple user have files to upload for RAG and/or transcription/translation
**When** those users concurrently upload their files via the UI or API
**Then** the files are queued up for sequential in-processing

### Describe alternatives you've considered
(optional) A clear and concise description of any alternative solutions or features you've considered.

### Additional context
In addition to queuing up multiple files uploaded by a single user, the queue system should also work across multiple users. For example, if several users are concurrently uploading audio files for translation (which can take a while), those files and the translation jobs should be queued up.

The queuing system should also provide a way to query/know where a file is within the queue, or if it is being in-processed, to enable the UI to provide feedback to users about the state of their upload.

There is a potentially related issue for [creating a queue/processing system that performs runs (#482)](https://github.com/defenseunicorns/leapfrogai/issues/482)

jalling97:
Completed by #942

PR #942, opened 2024-08-22
chore(api): add background task for processing vectors

### LINKED ISSUE #803: chore(api): Do not block when creating/updating VectorStores or Assistants
When hitting the endpoint to create or update VectorStores/Assistants there is a blocking operation that occurs with the indexing service. The implementation needs to be updated so that the indexing is run as a background job so that the user is not blocked waiting for the indexing to finish before being able to use an assistant.

### Additional context
[FastAPI background tasks](https://fastapi.tiangolo.com/tutorial/background-tasks/) are one way to do this. Indexing status is required before this can be done, as otherwise the frontend will be unable to determine when indexing has completed.

CollectiveUnicorn:
## Description

Configures the `api` and `text-embedings` to allow for embeddings and runs to be performed at the same time for a single `vectorstore` and `assistant` combination.

### CHANGES
* The indexing functions have been updated to create a background task when files are provided to it if a background task is provided.
* `text-embeddings` runs the embed function in an awaited thread to prevent blocking up to the grpc call.
* Updates test signup code to handle signup failures gracefully.
* Adds new e2e test that has llama-cpp-python, api, Supabase, and text-embeddings together.

## Related Issue

Fixes #803 #789

Relates to #626

## Checklist before merging


CollectiveUnicorn:
Blocked by #830 and the frontend integration with Supabase realtime for retrieving the VectorStore status

gphorvath (commented):
First pass, a few things jumped out.

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -236,43 +267,65 @@ async def modify_existing_vector_store(
...
+        )
+
+        if request.file_ids:
+            responses = await self.index_files(vector_store_id, request.file_ids)
+            for response in responses:
+                await self._increment_vector_store_file_status(vector_store, response)
+
+        vector_store.status = VectorStoreStatus.COMPLETED.value
+        last_active_at = int(time.time())
+        vector_store.last_active_at = last_active_at  # Update after indexing files
```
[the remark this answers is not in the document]
CollectiveUnicorn: I think its acceptable to use the code generated version here. As `last_active_time` will be used down below to determine expiry, we ideally don't want to have any drift between the value we used to determine expiry and the one that is set in the db.

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -89,7 +89,7 @@ async def index_file(self, vector_store_id: str, file_id: str) -> VectorStoreFil
                     ),
                     object="vector_store.file",
                     status=VectorStoreFileStatus.FAILED.value,
-                    usage_bytes=0,
+                    usage_bytes=0,  # Leave blank to have Postgres generate a UUID
```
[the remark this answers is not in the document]
CollectiveUnicorn: Yep, copypasta issue fixed now

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -100,7 +100,7 @@ async def index_file(self, vector_store_id: str, file_id: str) -> VectorStoreFil
                 last_error=None,
                 object="vector_store.file",
                 status=VectorStoreFileStatus.IN_PROGRESS.value,
-                usage_bytes=0,
+                usage_bytes=0,  # Leave blank to have Postgres generate a UUID
```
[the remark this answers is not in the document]
CollectiveUnicorn: Yep, copypasta issue fixed now

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -154,59 +154,90 @@ async def index_files(
...
+                name=request.name or "",
+                status=VectorStoreStatus.IN_PROGRESS.value,
+                object="vector_store",
                 created_at=0,  # Leave blank to have Postgres generate a timestamp
+                last_active_at=current_time,
                 file_counts=FileCounts(
                     cancelled=0, completed=0, failed=0, in_progress=0, total=0
                 ),
-                last_active_at=last_active_at,  # Set to current time
+                usage_bytes=0,  # Leave blank to have Postgres generate a UUID
```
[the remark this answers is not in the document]
CollectiveUnicorn: Yep, copypasta issue fixed now

tests/e2e/test_api.py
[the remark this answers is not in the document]
CollectiveUnicorn: This is a leftover from moving things around, removing the changes.

tests/e2e/test_text_backend_full.py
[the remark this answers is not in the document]
CollectiveUnicorn: Will await the PR being merged in to make changes.

CollectiveUnicorn:
Check for vector_store_file creation async

tests/e2e/test_text_backend_full.py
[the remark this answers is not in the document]
CollectiveUnicorn: I have updated the test to use the test fixture for the openai client.

src/leapfrogai_api/backend/rag/index.py
```diff
@@ -154,59 +154,90 @@ async def index_files(
...
                 expires_after=expires_after,
                 expires_at=expires_at,
             )
-            new_vector_store = await crud_vector_store.create(object_=vector_store)
 
-            if request.file_ids != []:
-                responses = await self.index_files(
-                    new_vector_store.id, request.file_ids
-                )
+            # Save the placeholder to the database
```
justinthelaw, line 185: I may have missed this, but at any point during the exception handling do we end up deleting the placeholder if the operation fails?

tests/e2e/test_text_backend_full.py
```diff
@@ -0,0 +1,160 @@
...
+import tempfile
+import requests
+import pytest
+from openai import OpenAI
+from openai.types.beta.vector_store import VectorStore
+
+from leapfrogai_api.backend.types import VectorStoreStatus
+
+
+def download_arxiv_pdf():
```
justinthelaw, line 11: To prevent the possibility of rate limiting, I recommend hosting or downloading this file somewhere else - not strictly required though!

tests/e2e/utils.py
```diff
@@ -20,15 +20,19 @@ def create_test_user(
...
+        requests.post(
+            url="https://supabase-kong.uds.dev/auth/v1/signup",
+            headers=headers,
+            json={
+                "email": email,
+                "password": password,
+                "confirmPassword": password,
+            },
+        )
+    except Exception:
```
justinthelaw, line 33: Recommend logging the exception in case something else happens besides a repeated user registration.

pyproject.toml
```diff
@@ -27,6 +27,7 @@ dev = [
     "requests-toolbelt",
     "pytest",
     "huggingface_hub[cli,hf_transfer]",
+    "fastapi",
```
justinthelaw, line 30: Is this being installed as a dev dependency for the test workflow? We should use the fastapi within the src/leapfrogai_api directory pyproject. Else, you can pin it here to match with that dep: `    "fastapi == 0.109.1",`

justinthelaw (changes requested):

pyproject.toml
CollectiveUnicorn: It doesn't use the api dependency, it uses the one at the root. Initially put it as a dev dependency on the API and it did not work, I'll pin the version here.

tests/e2e/test_text_backend_full.py
CollectiveUnicorn: I've used this PDF elsewhere and it was noted that we didn't want to keep large artifacts in the repo. So unless there's somewhere specific that you'd like it to be hosted I am going to leave this here for now.

justinthelaw: Makes sense! I know UDS hosts nginx in GHCR to prevent rate limiting, but since this is a straight unstructured file, that might not be an easy solution. I am not too worried about rate limiting, since we don't get as many pulls/runs as UDS does, but I thought I'd mention it for the chat history.

pyproject.toml
justinthelaw: Works for me!

tests/e2e/utils.py
CollectiveUnicorn: Yep I've added some logging that will explain that the error is likely due to the user already existing.

src/leapfrogai_api/backend/rag/index.py
CollectiveUnicorn: We don't, I've added some cleanup to the catch statement to delete the created placeholder.

Issue #944, opened 2024-08-23
(feat) Add Needle in a Haystack Evaluation Infrastructure


jalling97:
We should add the underlying infrastructure that runs the Needle in a Haystack evaluations that will ultimately be a part of the LFAI RAG Evals framework.

This does not include the code to run it in a pipeline. Just to be able to run the evaluation on a deployed instance of LFAI.

PR #945, opened 2024-08-23
feat: add RAG Evals Runner and Needle in a Haystack Evaluation

jalling97:
## Description
Adds a base RAG Evals runner and the Needle in a Haystack evaluation

### CHANGES
- Adds basic RAG Evals Python package in `src/`
- Adds RAGEvaluator class for running RAG Evals
- Adds runner for NIAH evaluation
- Adds README for background context and run details

## Related Issue
Relates to #944 

## Checklist before merging


evals/niah_runner.py
```diff
@@ -0,0 +1,260 @@
...
+                f"length: {row['context_length']}\n depth: {row['context_depth']}\n secret_code{row['secret_code']}"
+            )
+            # add file to vector_store
+            # TODO: there has to be a more efficient way to add this file
+            with open("context.txt", "wb") as context_file:
+                context_file.write(row["context"].encode("utf-8"))
+            with open("context.txt", "rb") as context_file:
+                vector_store_file = self.client.beta.vector_stores.files.upload(
+                    vector_store_id=self.vector_store.id, file=context_file
+                )
```
alekst23, line 68: Try this, it should convert your text directly to bytes, which should be an acceptable type for the `upload` method:
```suggestion
            vector_store_file = self.client.beta.vector_stores.files.upload(
                vector_store_id=self.vector_store.id, file=row["context"].encode('utf-8')
            )
```

evals/niah_runner.py
```diff
@@ -0,0 +1,260 @@
...
+                # retrieval_score
+                # 1 if needle text was returned by the retrieval step of RAG else 0
+                logging.info(
+                    f"number of annotations in response: {len(response.content[0].text.annotations)}"
+                )
+                for annotation in response.content[0].text.annotations:
+                    annotation_id = annotation.file_citation.file_id
+                    if annotation_id == file_id:
+                        logging.info("Setting retrieval_score to 1.0")
+                        retrieval_score = 1.0
```
alekst23, line 117: I think that this will always be true because we only upload one file to this assistant, and it will always be returned in the citation, even if the matching score is bad.

That might be why we are always getting 100% on this.

evals/niah_runner.py
```diff
@@ -0,0 +1,260 @@
...
+
+            # get messages
+            messages = self.client.beta.threads.messages.list(
+                thread_id=thread.id, run_id=run.id
+            ).data
+
+            response_messages = []
+            for message in messages:
+                if message.role == "assistant":
+                    response_messages.append(message)
```
alekst23, line 102: I am not sure if there will ever be more than one "assistant" message generated by the response. In this situation you can just test the last message in the list, which will be the new assistant generated message:
```
response_message = messages[-1]
```

jalling97: The only reason I did this is because sometimes the "user" message gets returned in the list after the "assistant" message. Not sure why this is but about half the time taking the last item in the list sometimes returned the "user" message.

evals/niah_runner.py
jalling97: The new test uploads (by default) 10 non-contextual documents as well, so it's not always a given.

jalling97:
Ultimately I think it makes the most sense to contain the evaluation framework within its own package. This helps with dependency management, and better aligns us for the long term needs of evals to be run within the deployment (e.g as a K8s job)

packages/evaluations/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+```bash
+# from within the packages/evaluations dir
+python -m pip install .
+python src/main.py
+```
+
+## Needle in a Haystack (NIAH)
+
+A Needle in a Haystack evaluation is used to evaluate the performance of LLMs in tasks that require finding a specific piece of information (the "needle") within a large body of text (the "haystack").
```
americanthinker, line 17: I would consider rewording this line to say that an "...evaluation is used to evaluate the performance of the LFAI RAG system in tasks...
Looks like we're measuring both retrieval (embedding model performance) and generation (LLM performance).

packages/evaluations/src/metrics/niah_metrics.py
```diff
@@ -0,0 +1,75 @@
...
+    def __init__(
+        self,
+        threshold: float = 0.75,
+        async_mode: bool = True,
+    ):
+        self.threshold = threshold
+        self.async_mode = async_mode
+
+    def measure(self, test_case: LLMTestCase):
+        """Records the niah retrieval score from the test case"""
```
americanthinker, line 18: What is being measured here and why is the threshold 0.75?  Isn't the needle either present or not in the retrieved documents?

americanthinker (commented):
Left two comments for clarification

CollectiveUnicorn (approved):

packages/evaluations/src/metrics/niah_metrics.py
jalling97: Good call out on the default threshold. You're right that we are simply measuring the presence of the retrieved context, which is a boolean operation. I updated the default threshold to 1.

Although at the moment, the metric itself doesn't actually do the measuring. That gets handled all within the niah_runner. In order for the niah tests to be run entirely on their own, the runner measures the presence of the needle in the retrieved context and the response and just reports that to the deepeval metrics directly.

packages/evaluations/README.md
jalling97: Made a quick update to the verbiage for clarity :+1:

packages/evaluations/src/runners/niah_runner.py
```diff
@@ -0,0 +1,361 @@
...
+        base_url: str = None,
+        api_key: str = None,
+        message_prompt: str = "What is Doug's secret code?",
+        min_doc_length: int = 4096,
+        max_doc_length: int = 4096,
+        min_depth: float = 0.0,
+        max_depth: float = 1.0,
+        num_copies: int = 2,
+    ):
+        """Initialize the Assistant with an API key and the path to the text file"""
```
gphorvath, line 65: Probably should expand this docstring to describe the parameters. I'm imagining this is reusable so I would like more clarity on padding, depth, doc length, and copies. Also it is probably important to mention that the prompt should specifically invoke a search for the needle.

packages/evaluations/src/runners/niah_runner.py
```diff
@@ -0,0 +1,361 @@
...
+                    )
+                padding_ids.append(padding_file.id)
+                os.remove(f"padding_{count}.txt")
+            logging.debug(
+                f"Added {len(self.padding)} files as padding to the haystack vector store"
+            )
+        self.vector_store = vector_store
+        self.padding = self.padding.add_column(name="padding_id", column=padding_ids)
+
+    def _delete_vector_store(self, vector_store_id: str) -> None:
```
gphorvath, line 348: Very minor nit, but this does more than delete the vector store - also deletes the files. Feel free to just resolve this if it doesn't help.

packages/evaluations/src/runners/niah_runner.py
```diff
@@ -0,0 +1,361 @@
...
+import matplotlib.pyplot as plt
+from tqdm import tqdm
+
+from matplotlib.colors import LinearSegmentedColormap
+from openai.types.beta.assistant import Assistant
+from openai.types.beta.vector_store import VectorStore
+
+load_dotenv()
+
+INSTRUCTION_TEMPLATE = """
```
gphorvath, line 19: should this be something the user can optionally override? I'm actually curious how much the assistant instructions affect it's ability to find the needle, after reading this: https://arize.com/blog-course/the-needle-in-a-haystack-test-evaluating-the-performance-of-llm-rag-systems/

packages/evaluations/src/metrics/niah_metrics.py
gphorvath: I think I'm still uncertain what `retrieval score` is. Is this accuracy?

packages/evaluations/README.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+You can then run the evaluation in a script with the following:
+
+```python
+from niah_runner import NIAH_Runner
+
+runner = NIAH_Runner() # see input parameters
+
+runner.evaluate() # run the evaluation
```
gphorvath, line 81: I think you renamed this to run_experiment()

packages/evaluations/src/runners/niah_runner.py
gphorvath, line 1: Only because I ran into this issue. It would be nice to pull out setup/cleanup from the `run_experiment` method. Basically I went through all the setup and then discovered I am running llama-cpp-python and vllm is the default, then had to run through it all over again, and I'm pretty sure the cleanup was never reached so I just have extra artifacts in my DB.

packages/evaluations/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+- LFAI is deployed
+- A valid LFAI API key is set
+
+Set the following environment variables:
+
+```bash
+LEAPFROGAI_API_URL=<lfai api url, usually: https://leapfrogai-api.uds.dev/openai/v1 for development>
+LEAPFROGAI_API_KEY=<lfai api key>
+```
```
gphorvath, line 72: Might be good to expose a variable for model, to make main more flexible. This way I can easily change to openai/gpt4o, lfai/llama-cpp-python, etc.

gphorvath (commented):
Looks like a great start! I have a bunch of minor comments, take what you find helpful and resolve the rest.

packages/evaluations/src/metrics/niah_metrics.py
```diff
@@ -0,0 +1,75 @@
...
+            self.reason = f"Retrieval in the NIAH evaluation scored less than the threshold score of {self.threshold}"
+
+        return self.score
+
+    async def a_measure(self, test_case: LLMTestCase):
+        loop = asyncio.get_running_loop()
+        return await loop.run_in_executor(None, self.measure, test_case)
+
+    def is_successful(self):
+        return self.success
```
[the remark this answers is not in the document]
jalling97: This is required under the hood for DeepEval: https://docs.confident-ai.com/docs/metrics-custom#4-implement-the-is_successful-method

packages/evaluations/src/metrics/niah_metrics.py
```diff
@@ -0,0 +1,75 @@
...
+            self.reason = f"Response in the NIAH evaluation scored less than the threshold score of {self.threshold}"
+
+        return self.score
+
+    async def a_measure(self, test_case: LLMTestCase):
+        loop = asyncio.get_running_loop()
+        return await loop.run_in_executor(None, self.measure, test_case)
+
+    def is_successful(self):
+        return self.success
```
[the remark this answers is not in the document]
jalling97: Same as above, this method is required by DeepEval for all metrics: https://docs.confident-ai.com/docs/metrics-custom#4-implement-the-is_successful-method

packages/evaluations/src/runners/niah_runner.py
jalling97: For sure! I went down a rabbit hole of attempting to identify all the parameters that we may want to tune on via some kind of hyperparameter search, and this is one of the parameters I missed. I'll add it.

packages/evaluations/README.md
jalling97: Good catch, I'll rename

packages/evaluations/src/metrics/niah_metrics.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __init__(
+        self,
+        threshold: float = 1.0,
+        async_mode: bool = True,
+    ):
+        self.threshold = threshold
+        self.async_mode = async_mode
+
+    def measure(self, test_case: LLMTestCase):
```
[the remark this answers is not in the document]
jalling97: added :+1:

packages/evaluations/src/metrics/niah_metrics.py
```diff
@@ -0,0 +1,75 @@
...
+        """Records the niah retrieval score from the test case"""
```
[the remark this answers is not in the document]
jalling97: expanded :+1:

packages/evaluations/src/runners/niah_runner.py
jalling97: Added an override :+1:

packages/evaluations/README.md
jalling97: Added :+1:

packages/evaluations/README.md
```diff
@@ -0,0 +1,85 @@
...
+The LeapfrogAI NIAH evaluation assumes the following:
+
+- LeapfrogAI is deployed
+- A valid LeapfrogAI API key is set
+
+Set the following environment variables:
+
+```bash
+LEAPFROGAI_API_URL=<LeapfrogAI API url, usually: https://leapfrogai-api.uds.dev/openai/v1 for development>
+LEAPFROGAI_API_KEY=<LeapfrogAI API key>
```
andrewrisse, line 71: Should we add a note about how/where to get a LEAPFROGAI_API_KEY? Or is this discussed somewhere else

packages/evaluations/src/runners/niah_runner.py
jalling97: I pulled out the cleanup steps into its own function that runs when the experiment is done or if an error occurs :+1:

packages/evaluations/README.md
jalling97: The API README covers this so I can link to it in the README

jalling97: linked :+1:

packages/evaluations/src/metrics/niah_metrics.py
jalling97: `retrieval score` is effectively a boolean that measures whether or not the `secret code` was found in the context that gets retrieved from RAG. This ends up getting averaged across all the runs for an aggregate score. Unfortunately we don't have an easy way of validating this at the moment, so I'm using the proxy of verifying that the file that contains the secret word is found in the annotations. It's possible that RAG could return a chunk from the right file that doesn't actually contain the needle, so this approach is slightly flawed until we can return the full retrieved context back to the user.

gphorvath (dismissed):
Awesome, looks like you hit all my comments! Looks good to me.

CollectiveUnicorn (dismissed):

justinthelaw (dismissed):

gphorvath (approved):
Move to `src/` looks good to me!

src/leapfrogai_evals/main.py
```diff
@@ -0,0 +1,70 @@
...
+            logging.info("Setting eval list to ALL")
+            self.eval_list = ALL_EVALS
+        # TODO: Add other evals options
+
+    def run_evals(self, *args, **kwargs) -> None:
+        """Run all of the selected evaluations"""
+        logging.info("Running the following evaluations:")
+        for eval in self.eval_list:
+            logging.info(f" -{eval}")
+        if "LFAI_NIAH" in self.eval_list:
```
vanakema, line 32: Python has a variety of ways you can do this to where you don't have to do an if statement here, but instead call it dynamically. Here's a couple options that might simplify your code to prevent you from having to write more if statements as you add more evals, keeping the code more concise:

### Using getattr
```
class MyClass:
    def function_one(self):
        print("Function One Called")

    def function_two(self):
        print("Function Two Called")

    def function_three(self):
        print("Function Three Called")

# Create an instance of the class
my_object = MyClass()

# List of method names as strings
method_names = ['function_one', 'function_two', 'function_three']

# Loop through the list and call each method by name
for name in method_names:
    method = getattr(my_object, name)  # Get the method from the object
    method()  # Call the method
```

**Or...** if you wanna get **really leet**, you can make it so you don't even have to define a list of functions at all, but instead dynamically load evals defined in a directory of files, each of which contains its own eval. Personally I think this is the cleanest way of doing it, and all anyone needs to do in the future is create a new `eval_*.py` file and it will automatically work. Would make it easy for people coming to the repo to fork or to contribute.

### The dynamic loading example thingy
Imagine the directory structure looks like such:
```
packages/
└── evaluations/
    └── src/
        ├── base_eval.py
        ├── main.py
        └── evals/
            ├── eval_1.py
            ├── eval_2.py
            └── eval_3.py
```
Here's what `base_class.py` would look like
```
from abc import ABC, abstractmethod

class BaseEval(ABC):
    @abstractmethod
    def run_eval(self):
        raise Error("Not implemented yet")
```
Here's what `eval_1.py` and `eval_2.py` would look like
```
# eval_1.py
from base_eval import BaseEval

class Eval1(BaseEval):
    def run_eval(self):
        print("Eval1 running evaluation")

# eval_2.py
from base_eval import BaseEval

class Eval2(BaseEval):
    def run_eval(self):
        print("Eval2 running evaluation")
```

And here's the logic in `main.py`
```
import os
import importlib
import inspect
from base_eval import BaseEval

evals_dir = './evals'

# Loop through each file in the directory
for filename in os.listdir(evals_dir):
    if filename.endswith(".py"):
        module_name = filename[:-3]  # Strip the .py extension
        module_path = f'{evals_dir.replace("/", ".")}.{module_name}'
        module = importlib.import_module(module_path)

        # Inspect the module for classes
        for name, obj in inspect.getmembers(module, inspect.isclass):
            # Check if it's a subclass of BaseEval and not BaseEval itself
            if issubclass(obj, BaseEval) and obj is not BaseEval:
                instance = obj()  # Instantiate the class
                instance.run_eval()  # Call the run_eval method
```

Personally I think everything you need to do your eval should be explicitly defined in the `run_eval()` function signature, BUT if you think that's going to differ too often, you can always pass in `self` into the function from `main.py` so it has all the same vars available to it.

We used this at my last job as a pattern, and it was really convenient

vanakema: You may be able to use this too with your metrics too, but I haven't looked hard enough at it to tell

vanakema (commented):
Looks pretty good. Left some suggestions. Only other suggestion I'd add as for the code itself is to maybe move this stuff into `/test` instead of `/src` since it's its own independent thing that's meant for evaluating the system as a whole, cuz then you could set it up so these evals also run via pytest and have all the env vars set to what they should be for when they're run locally vs in CI, etc.

To me, code in `src` should be code that's involved in the actual running of services. If the plan is to expose that via the API, then maybe it does deserve to be in `src`, and then for making it easy to run with the rest of your tests, you could just call it from your tests, and use one of the asserts that looks for whether the score is within the threshold.

vanakema:
I also have some comments on the dataset being used to eval too:
It seems like the needle in the NIAH dataset is in great contrast to the surrounding context. I think type of "needle being very specific" dataset is probably good for LLM's in terms of testing attention, but idk how useful it is in terms of actually finding stuff via embedding search. Documents in an embedding database are going to likely be somewhat related. At least they'll be in the domain, and the contrast with the rest of the documents will be less severe, so the embeddings are bound to be closer in cosine distance than a real world case.

Would love to help with it. I'm thinking we can use a dataset of real unclass gov documents and are all surrounding a specific topic, such as submarines. This would be more realistic to what we would be asking for from the system. I wonder if having it be actual output from pdfs, since the formatting that we get back from Unstructured is all borked up, would be good too.

Unrelated: the secret code phrase in the row where the secret_code = Hotel117 is actually concatenated to another sentence with no space or anything separating them, which makes it have no demarcator from the rest of it.

CollectiveUnicorn (approved):

src/leapfrogai_evals/main.py
jalling97: Great options! I mostly went with the list in the first place so that it would be easier for the user to decide which evals they want to run (just provide a list and the runner handles the rest), but I do like the dynamic loading option. 

I think it makes sense to refactor this with the next [draft PR](https://github.com/defenseunicorns/leapfrogai/pull/960) that adds new evals, so it'll be more clear how to implement having multiple evals in that one.

jalling97: As for where these go, we discussed in PR review and landed on `src/` as a place for the evals to live because they're going to operate differently from our tests. Evaluations are going to be run against a deployed instance of LFAI (whether that's local or not) since we intend to add performance evals at some point. Evaluations are going to be run in CI as a nightly/weekly workflow, and likely not on every PR unless it's specifically one that impacts the system performance. Since we don't want to mock anything for evals, and the evaluations will likely need to be a deployable K8s job, keeping it in `src/` had a few more pros than coupling it with the tests.

jalling97:
@vanakema in regards to the dataset, I'd love to work with you to get a dataset using more representative context. We made this dataset to mimic how other needle-in-a-haystacks tests tend to operate, but I agree that in this case the needle may be too high-contrasting to be as useful as it can be. Happy to start talking about that with you in more detail.

PR #951, opened 2024-08-26
feat(ui): chat completion with files

andrewrisse:
## Description
Feature enables uploading files for chat when not using an assistant from the chat page.
The files are converted to pdf if needed, then parsed for text. The text gets stored as it's own message with metadata that indicates the message should be hidden (on the UI).

## Checklist before merging

https://www.loom.com/share/9e9a1cb87217454f9e0fd9db0436b9fe?sid=9ecb91a7-aa82-4d2f-8acc-20da2bfb61c2

Deployed on LF-04


gregclark-defenseunicorns (commented):
@andrewrisse 

Found one thing I wanted address. Seems like the submit button for the prompt message is still enabled when a file is uploading to the message and making a weird edge case.

Steps:
- Upload a file to the message
- Submit the message while the file is uploading
- File in message thread will show as uploading permanently
- The File will then upload to the message bar without having to upload it again.

It would be best to disable the submit message button while the file is uploading to avoid this scenario

andrewrisse:
> @andrewrisse
> 
> Found one thing I wanted address. Seems like the submit button for the prompt message is still enabled when a file is uploading to the message and making a weird edge case.
> 
> Steps:
> 
> * Upload a file to the message
> * Submit the message while the file is uploading
> * File in message thread will show as uploading permanently
> * The File will then upload to the message bar without having to upload it again.
> 
> It would be best to disable the submit message button while the file is uploading to avoid this scenario
> 
>  Screen.Recording.2024-08-29.at.1.01.07.PM.mov

fixed

justinthelaw (approved):

PR #953, opened 2024-08-27
test(e2e): move openai client to fixture

alekst23:
## Description

Test discovery in IDE can fail due to code that runs on dependency resolution. This code makes web requests to retrieve access tokens, but can fail due to a number of reasons. As a result, test discovery will fail and the IDE will not show correct information about your test functions, and will have many errors in the test Output log, making debugging other issues more difficult.

### CHANGES

This PR moves OpenAI client into a fixture that will only run once for each module. This will prevent the IDE from throwing errors on discovery.

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -1,0 +1,12 @@
+import pytest
+
+from openai import OpenAI
+
+from .utils import create_test_user
+
+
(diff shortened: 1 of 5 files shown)


jalling97 (approved):
Just needs a label added and it looks good!

gphorvath (approved):

PR #958, opened 2024-08-28
chore: pin python dependencies in root pyproject

YrrepNoj:
This PR pins the dependencies in the root pyproject.toml. This toml file is used in some of our workflows (such as e2e testing and release building) so it is important to have these deps pinned as we have for our other pyproject.tomls.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ readme = "README.md"
 license = { file = "LICENSE" }
 
 dependencies = [ # Dev dependencies needed for all of LeapfrogAI
-    "openai == 1.31.1",
+    "openai == 1.32.1",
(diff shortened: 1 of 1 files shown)


justinthelaw (approved):

gphorvath (approved):

pyproject.toml
```diff
@@ -11,11 +11,11 @@ readme = "README.md"
 license = { file = "LICENSE" }
 
 dependencies = [ # Dev dependencies needed for all of LeapfrogAI
-    "openai",
+    "openai == 1.32.1",
```
alekst23, line 14: Why do we pin openai to this version? Latest version is 1.43.0. We already have some non-conformance. Is there specific features we are trying to preserve in their state?

Issue #948, opened 2024-08-26
chore: Organize code into Types, Models, and a Composer


alekst23:
# Context

Our API response handling code is not very uniform or modular. It uses several methods, interchangeably, to perform operations on Database and LLM. Some of the code imports router functions instead of modules to perform tasks like completion.

# Description

- We should create a system that uses repeatable patterns to handle router logic and perform system tasks like DB calls and LLM operations.
- We should create a centralized module to handle LLM features like message composition, RAG, tool calling, etc.

https://drive.google.com/file/d/1dWVcqctJbyVx0hWP6XE84kMpGYICCkic/view?usp=sharing

alekst23:
regarding Usage class:
https://github.com/defenseunicorns/leapfrogai/pull/963#discussion_r1746219829

PR #963, opened 2024-08-28
chore: new 'types' module

alekst23:
## Description
We need a better way to organize our type definitions. Currently we have the following issues:
- One file with a lot of definitions: `src/leapfrogai_api/backend/types.py`. Very long file with many types for different purposes defined together. Difficult to analyze and find code.
- Some types mix data models and complex function implementations in one class
- Types are not defined in a consistent manner

Good format to follow is the OpenAI types module:
https://github.com/openai/openai-python/tree/main/src/openai/types

### BREAKING CHANGES
All changes to type definitions should be applied to all relevant files as part of the PR.

### CHANGES
- A new module for `leapfrogai_api` is created: `types`.
- Python Module format allows for complex type hierarchy and selective importing of objects. 
- Creates a standardized way for application to define and use types.
- Consolidates types defined inline and in `src/leapfrogai_api/backend/types` to the new `types` module
- Consolidates constants to `src/leapfrogai_api/types/constants.py`

## Related Issue

Precursor to: 
https://github.com/defenseunicorns/leapfrogai/issues/948

## Checklist before merging

- [ ] Followed the [Contributor Guide Steps](https://github.com/defenseunicorns/leapfrogai/blob/main/.github/CONTRIBUTING.md)


CollectiveUnicorn (approved):

src/leapfrogai_api/backend/thread_runner.py
gphorvath, line 1: propose we rename this to not confusion OpenAI Threads with Threading Threads

src/leapfrogai_api/typedef/constants.py
```diff
@@ -0,0 +1,6 @@
+# This file defines application constants not set by environment variables or configuration files.
+
+THIRTY_DAYS = 60 * 60 * 24 * 30  # in seconds
```
gphorvath, line 3: As this travels farther and farther from where it was first defined/use, I think it is worth updating the variable name to `THIRTY_DAY_IN_SECONDS` or something else that preserves the units.

src/leapfrogai_api/typedef/constants.py
gphorvath, line 1: I'm not sure I'm a fan of having constants at the root of a types directory. Perhaps `leapfrogai_api/_constants.py`?

src/leapfrogai_api/typedef/common.py
gphorvath, line 1: I wonder if Usage is still necessary or if we could switch to OpenAI's completion_usage type to eliminate the need for this file?

https://github.com/openai/openai-python/blob/main/src/openai/types/completion_usage.py

src/leapfrogai_api/backend/thread_runner.py
alekst23: Renaming to Composer

src/leapfrogai_api/typedef/constants.py
alekst23: I would like to avoid putting things in root, as it will lead to accumulation of gunk in our project root directory after a while. How about `leapfrogai_api/backend/_constants.py`? And we can set up the import via the backend module like to work like this: `from leapfrogai_api.backend import CONSTANT_VALUE`. How does that sound?

src/leapfrogai_api/typedef/common.py
alekst23: To make matters more complicated, it appears that we have a triplicate:
`src/leapfrogai_sdk/completion/completion_pb2.pyi` also defined `class CompletionUsage`

CollectiveUnicorn (approved):

src/leapfrogai_api/typedef/constants.py
alekst23: Done

PR #971, opened 2024-08-30
chore: update CODEOWNERS

daveworth:
### CHANGES

Replace Austen Bryan and Gerred with Dave Worth as CODEOWNERS

## Checklist before merging

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,5 +2,5 @@
 * @defenseunicorns/leapfrogai
 
 # Additional privileged files
-/CODEOWNERS @jeff-mccoy @austenbryan @gerred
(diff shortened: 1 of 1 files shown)


justinthelaw:
Nit-pick! as part of this change, can we move CODEOWNERS into the `.github/` directory to lessen the current (and soon more) root-level file count? Just to clean-up the root-level directory some more, since we have a lot there.

daveworth:
I think that would generally be different from the majority of projects in the Defense Unicorns org - I will defer to @YrrepNoj on this suggestion.

YrrepNoj:
While I love a clean root directory I think CODEOWNERS is an important file and should stay at the root 🫶

gerred (approved):
yay lgtm!

PR #974, opened 2024-09-03
fix: correct fsGroup configuration in various deployments

YrrepNoj:
Several (if not all 😬) of our deployments had the `fsGroup` configuration in the wrong place so it wasn't actually applying. This led to an issue where our data-injections were not working when the underlying store was not a PVC provisioned with `local-path` as the StorageClass.

This PR fixes the fsGroup key so that it is in the correct place.

### DIFF
--- a/packages/api/chart/templates/api/deployment.yaml
+++ b/packages/api/chart/templates/api/deployment.yaml
@@ -72,7 +72,6 @@ spec:
           securityContext:
             runAsUser: 65532
             runAsGroup: 65532
-            fsGroup: 65532
           volumeMounts:
           - name: api-model
             mountPath: /config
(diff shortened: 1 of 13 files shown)


justinthelaw (commented):
Awesome! I also [will have] fixed this [here](https://github.com/defenseunicorns/leapfrogai/pull/937). This works locally, I but am not sure if this actually helps in a CSP deployment as well.

Have you tested this in an EC2-based cluster using EFS yet?

justinthelaw (approved):
Post-standup context approve!

YrrepNoj:
> Have you tested this in an EC2-based cluster using EFS yet?

Yes! We replicated and validated this fix on an RKE2 cluster (running on an EC2 instance) that was using `EBS` as the StorageClass.

CollectiveUnicorn (approved):

PR #978, opened 2024-09-03
fix(ui): delete avatar

andrewrisse:
## Description
When removing an avatar from an assistant, the sveltekit backend was not actually deleting the avatar image from Supabase storage. 

### CHANGES

Corrects the file path and adds a Row Level Security policy that allows users to delete their own avatars.

The 0.7.2 migration that was modified was edited because there was a duplicate policy.

## Checklist before merging


CollectiveUnicorn (approved):

justinthelaw (dismissed):

src/leapfrogai_ui/supabase/migrations/20240322174521_v0.7.2_ui_sql_schema.sql
```diff
@@ -70,12 +70,6 @@ BEGIN
           for update using (auth.uid() = owner) with check (bucket_id = 'avatars');
     END IF;
 
-    IF NOT EXISTS(SELECT 1 FROM pg_policies WHERE schemaname = 'storage'
```
YrrepNoj, line 73: In general, we should avoid modifying existing migration files whenever possible. We've broken this rule a few times in the past but I would like to avoid breaking it as much as we can. Can you return this block back to the way it was?

src/leapfrogai_ui/supabase/migrations/20240903204301_v0.12.0_avatars.sql
```diff
@@ -0,0 +1,73 @@
...
+
+    IF NOT EXISTS(SELECT 1 FROM pg_policies WHERE schemaname = 'storage'
+      AND tablename  = 'objects'
+      AND policyname = 'Anyone can update their own assistant avatars.')
+        THEN create policy "Anyone can update their own assistant avatars." on storage.objects
+          for update using (owner = auth.uid() AND bucket_id = 'assistant_avatars');
+    END IF;
+
+
+    IF NOT EXISTS(SELECT 1 FROM pg_policies WHERE schemaname = 'storage'
```
YrrepNoj, line 32: (here and above on line 24): Do we want to 'error' if a policy already has this name?

Migration scripts are only one once so if a database already had a policy with that name (either the drop didn't work or someone was tinkering and added it themselves) then these policies would never be applied.

src/leapfrogai_ui/supabase/migrations/20240322174521_v0.7.2_ui_sql_schema.sql
YrrepNoj: Also, what was the end goal of removing this policy here?

If someone has an app that was running with a pre v0.12.0 deployment then they would have this policy. Removing these lines would not remove the policy when they updated to a post v0.12.0 deployment. If this policy is no longer relevant we will need to add a `DROP POLICY IF EXISTS` for it.

src/leapfrogai_ui/supabase/migrations/20240903204301_v0.12.0_avatars.sql
YrrepNoj: Don't think there's a wrong answer for this one, I am mostly pondering how we want error flow to be.

YrrepNoj (changes requested):
One general question and one change request 🫶

src/leapfrogai_ui/supabase/migrations/20240322174521_v0.7.2_ui_sql_schema.sql
andrewrisse: It was a duplicate policy, so the policy will still exist, it was in this file twice

andrewrisse: and the first occurence was mixed with policies for the ol "avatar" bucket, which was misleading

src/leapfrogai_ui/supabase/migrations/20240903204301_v0.12.0_avatars.sql
andrewrisse: hmm, currently it just follow existing patterns. I remember we added stuff like this initially because we were getting lots of errors, but you are right that if applied correctly, we wouldn't need it. I guess it helps when fixing issues if you are copy pasting migrations manually

andrewrisse:
@YrrepNoj , Gato has the migration in this PR: https://github.com/defenseunicorns/leapfrogai/pull/985
tagged as v0.11.1, should I update this PR to match?

src/leapfrogai_ui/supabase/migrations/20240322174521_v0.7.2_ui_sql_schema.sql
andrewrisse: removed that change

CollectiveUnicorn (dismissed):

jalling97 (dismissed):

src/leapfrogai_ui/supabase/migrations/20240322174521_v0.7.2_ui_sql_schema.sql
andrewrisse: @YrrepNoj following up on this and want to make sure we are good with your changes requested

YrrepNoj (approved):

PR #988, opened 2024-09-05
chore: update release workflow trigger type

YrrepNoj:
This PR updates our release workflow so that the `tag-for-release` workflow triggers the `publish` workflow (instead of the `publish` workflow being triggered by a tag push.)

Motivation: When the GitHub Bot would create a new tag in our repository the `on.push.tag` trigger would not actually trigger and a manual tag push was required. This change streamlines our release workflow to not require that middle step.

The trigger logic for this change was heavily inspired (:cough: copied :cough:) from [uds-core/.github/workflows/tag-and-release.yaml](https://github.com/defenseunicorns/uds-core/blob/main/.github/workflows/tag-and-release.yaml) and [uds-core/.github/workflows/publish.yaml](https://github.com/defenseunicorns/uds-core/blob/main/.github/workflows/publish.yaml)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -38,7 +38,7 @@ jobs:
         id: get_version
         run: |
           LFAI_VERSION=$(jq -r '.["."]' .github/.release-please-manifest.json)
-          echo "LFAI_VERSION=$LFAI_VERSION" >> $GITHUB_ENV
+          echo "LFAI_VERSION=$LFAI_VERSION" >> $GITHUB_OUTPUT
 
       - name: Instal Python Deps
         run: |
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,9 +1,7 @@
 name: Publish Release Artifacts
 
 on:
-  push:
-    tags:
-      - "*"
+  workflow_call
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -38,7 +32,13 @@ jobs:
...
         uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c #v5.0.0
         with:
-          python-version-file: 'pyproject.toml'
+          python-version-file: "pyproject.toml"
+
+      - name: Get Version
+        id: get_version
+        run: |
+          LFAI_VERSION=$(jq -r '.["."]' .github/.release-please-manifest.json)
+          echo "LFAI_VERSION=$LFAI_VERSION" >> $GITHUB_ENV
```
justinthelaw, line 41: This will not work because $GITHUB_ENV requires `echo "::set-output name=LFAI_VERSION::$LFAI_VERSION"` to be accessible elsewhere using the `{{ steps.get_version.outputs.LFAI_VERSION }}` method.

You could also, instead of the above, use this: `"LFAI_VERSION=$LFAI_VERSION" >> $GITHUB_OUTPUT`, since `"::set-output ` is deprecating.

See working example: https://github.com/defenseunicorns/leapfrogai/blob/eee3ed7b570d728cc10a2e4a9dec6bfdb38de2c4/.github/workflows/e2e-registry1-weekly.yaml#L76

Also see line 23 in this PR's `tag-for-release.yaml` for a working example too!

justinthelaw (changes requested):

.github/workflows/release.yaml
YrrepNoj: Good catch, I meant to pipe this to `GITHUB_OUTPUT` and not `GITHUB_ENV`. How does this look now?

justinthelaw (approved):
Everything looks good and seems to work as independent components. I can _try_ to break this down into 1 artifact being published (I don't currently have large, personal runners) in a fork of the repo and run it tomorrow morning to confirm.

CollectiveUnicorn (approved):

CollectiveUnicorn:
I don't see any issues with the workflow as written, the conventions all seem correct

Issue #994, opened 2024-09-06
bug: chatting with an assistant without files fails silently


justinthelaw:
### Environment

1. OS and Architecture: Ubuntu 22.04 AMD64
2. App or Package Name: leapfrogai-api
2. App or Package Version: main
3. Kubernetes Distribution: uds-k3d
4. Kubernetes Version: 1.28.8
5. Other: N/A

### Steps to reproduce

1. Create an assistant without a file attached in the UI
2. Chat with the assistant

### Expected result

- Chat message is returned from the run

### Actual Result

- The run fails as the API router's request builder expects a `file_ids` to not be empty

### Visual Proof (screenshots, videos, text, etc)

```bash
Traceback (most recent call last):                                                                                                                               
  File "/leapfrogai/.venv/lib/python3.11/site-packages/starlette/responses.py", line 257, in wrap                                                                
    await func()                                                                                                                                                 
  File "/leapfrogai/.venv/lib/python3.11/site-packages/starlette/responses.py", line 246, in stream_response                                                     
    async for chunk in self.body_iterator:                                                                                                                       
  File "/leapfrogai/.venv/lib/python3.11/site-packages/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py", line 380, in stream_generate_message_for_thread                                                                                                                                                      
    chat_messages, file_ids = await self.create_chat_messages(                                                                                                   
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                   
  File "/leapfrogai/.venv/lib/python3.11/site-packages/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py", line 286, in create_chat_messages                                                                                                                                                                    
    return chat_messages, list(file_ids)                                                                                                                         
                               ^^^^^^^^                                                                                                                          
UnboundLocalError: cannot access local variable 'file_ids' where it is not associated with a value 
```

### Additional Context

N/A

Issue #564, opened 2024-05-28
test: Convert integration tests into e2e test and add to pipeline


CollectiveUnicorn:
### Describe what should be investigated or refactored
Convert the integration tests in `tests/integration` into e2e tests and move them into the e2e test pipeline.

PR #1004, opened 2024-09-06
chore(test): fix and re-activate integration and unit tests

### LINKED ISSUE #1001: chore(test): write backend integration test for PR #995

See https://github.com/defenseunicorns/leapfrogai/pull/995 for details.

https://github.com/defenseunicorns/leapfrogai/blob/main/src/leapfrogai_api/routers/openai/requests/run_create_params_request_base.py

### Additional context

Should cover the case where a run on a thread has an assistant that has access to no tools.

justinthelaw:
## Description

See PR #994 for details on testing issues, and #564 for the overall integration testing issues. If the integration tests were run prior to committing, or in the pipeline, the `file_ids` issue fixed by PR #994 (related to Issue #1001) would have been caught earlier.

### BREAKING CHANGES

- N/A

### CHANGES

- Fixes the Makefile targets for executing integration tests - now works on all shells and scenarios
- Further improves README for local unit and integration test execution
- Enables integration and unit testing in local development and inside our CI pipelines
- Refactors the Repeater to better mock an actual in-cluster model, fixing issues with chat completion integration tests and streaming chat unit tests
- Refactors unit tests to test the behavior of the API and SDK, rather than the Repeater implementation code
- Bumps `unstructured` and `nltk` versions to allow for `.pptx` chunking, which was previously not working and failed integration tests

## Related Issue

Fixes Issues #1001 #564
Related to PR #994

## Checklist before merging


jalling97 (approved):
Having the integration tests be a part of the pipeline is fantastic!

Just pointing out though that the `pytest` workflow takes about 25 minutes now, which is about as long as it takes to test `text-embeddings`. So while we're not taking any longer to run all of the tests, we're further cementing that 25-30 minute timeframe. If we make changes in the future to futher optimize the E2E tests, addressing the integration tests will probably also be necessary.

justinthelaw:
> Having the integration tests be a part of the pipeline is fantastic!
> 
> Just pointing out though that the `pytest` workflow takes about 25 minutes now, which is about as long as it takes to test `text-embeddings`. So while we're not taking any longer to run all of the tests, we're further cementing that 25-30 minute timeframe. If we make changes in the future to futher optimize the E2E tests, addressing the integration tests will probably also be necessary.

@jalling97 totally agree! I will make an evergreen spike to investigate speeding up all e2e tests, including integration tests. The difficulty is definitely that each backend is very time-consuming to install and build. What we can do in the spike is investigate GH caching (like I do with the pip deps) and also more usage of a no/low-weight model like the repeater or T5, GPT2 etc., in tests-only.

jalling97 (approved):

CollectiveUnicorn (approved):

PR #1006, opened 2024-09-06
chore(ui): html/markdown security

andrewrisse:
## Description
Updates the order of sanitization of text rendered in messages to improve security.

## Checklist before merging


src/leapfrogai_ui/src/lib/components/Message.svelte
```diff
@@ -178,7 +178,13 @@
...
             {:else}
               <!--eslint-disable-next-line svelte/no-at-html-tags -- We use DomPurity to sanitize the code snippet-->
-              {@html md.render(DOMPurify.sanitize(messageText))}
+              {@html DOMPurify.sanitize(md.render(messageText), {
+                CUSTOM_ELEMENT_HANDLING: {
+                  tagNameCheck: (tagName) => tagName.match(/code-block/),
+                  attributeNameCheck: (attr) => attr.match(/code/) || attr.match(/language/),
+                  allowCustomizedBuiltInElements: false
+                }
+              })}
```
bburky, line 187: This looks good. Thanks for reaching out so we can walk through this.

As I mentioned, you can use a regex object instead of an function for `tagNameCheck` and `attributeNameCheck` if you want. Either way is fine though.

Please do make sure to match the whole string though if you use a regex, the regexes you currently have would also match `code-block-something-else` which isn't desired.

I haven't tested this, but I think this should work:
```suggestion
              {@html DOMPurify.sanitize(md.render(messageText), {
                CUSTOM_ELEMENT_HANDLING: {
                  tagNameCheck: /^code-block$/,
                  attributeNameCheck: /^(code|language)$/,
                  allowCustomizedBuiltInElements: false
                }
              })}
```

andrewrisse: updated, thanks!

bburky (dismissed):
I only really reviewed src/leapfrogai_ui/src/lib/components/Message.svelte, but the change looks good to me.

I'll review CodeBlock comprehensively sometime in the future, but if that extra sanitization doesn't break anything, that's good to add.

justinthelaw (approved):

gphorvath (approved):

PR #1029, opened 2024-09-12
chore: split up make clean target

gphorvath:
## Description

Splitting up some of the make targets for cleaning up artifacts, cache, etc. This will help when the intent is only delete build artifacts or models, without also hitting .env or ruff_cache.

### BREAKING CHANGES

None

### CHANGES

* Replaced `make clean` with an include for `mk-clean.mk`
* Split `make clean` functionality into `clean-artifacts`, `clean-cache`, `clean-env`, `clean-models`, and `clean-logs`
* Added `make clean-all` to replicate previous functionality.

## Related Issue

None, just an problem spotted in the course of rebuilding.

## Checklist before merging


justinthelaw (approved):
One nitpick that technically doesn't need to be here - can we include a UDS clean too?

This works on Debian based distros, but the temp files may be located elsewhere in Mac and Windows. The temp files do take up a TON of space if your OS clean-up job isn't often or thorough.

```bash
uds zarf tools clear-cache && rm -rf ~/.uds-cache && rm -rf /tmp/zarf-*
```

gphorvath:
> One nitpick that technically doesn't need to be here - can we include a UDS clean too?
> 
> This works on Debian based distros, but the temp files may be located elsewhere in Mac and Windows. The temp files do take up a TON of space if your OS clean-up job isn't often or thorough.
> 
> ```shell
> uds zarf tools clear-cache && rm -rf ~/.uds-cache && rm -rf /tmp/zarf-*
> ```

@justinthelaw I don't disagree that this is a valuable alias. My only concern with adding it to this PR is that it causes our make clean to exceed the scope of LeapfrogAI. By limiting our cleanup to files within this directory we avoid that. We have `uds-clean` documented as an alias in `DEVELOPMENT.md`, do you think that is sufficient to satisfy your need?

justinthelaw (approved):

PR #1031, opened 2024-09-13
fix(sdk): best_of in completions was an int and should be a string

### LINKED ISSUE #819: bug(api): Fix legacy `completions` API endpoint
The legacy `/openai/v1/completions` endpoint is broken and returns an `500` Internal Server Error. This endpoint should be fixed to operate according to the legacy endpoint specification.

- [Legacy `completions` endpoints](https://platform.openai.com/docs/api-reference/completions/create)
- [New `chat/completions` endpoint](https://platform.openai.com/docs/api-reference/chat/create)

### Additional context
The `Continue.dev` plug-in seems to require this endpoint, so we should continue to maintain it until we learn otherwise.

lemonprogis:
## Description

the `best_of` in completion.proto hasn't been changed to a string like in others. Changed that and ran the code gen. 

### BREAKING CHANGES
N/A

## Related Issue

Fixes #819 

## Checklist before merging

- [ ] Followed the [Contributor Guide Steps](https://github.com/defenseunicorns/leapfrogai/blob/main/.github/CONTRIBUTING.md)


src/leapfrogai_sdk/audio/audio_pb2.py
gphorvath, line 1: It appears the lint is change tab size from 4 to 2 spaces. 4 should be the default for Python

src/leapfrogai_sdk/audio/audio_pb2.pyi
gphorvath, line 1: In this file it seems like somebody's linter config is undoing the line length limit and removing trailing whitespace.

src/leapfrogai_sdk/audio/audio_pb2_grpc.py
gphorvath, line 1: Double quotes changing to single quotes - check linter config.

gphorvath (changes requested):
I would like to clear up the linting issues before merging. There is a lot of this that is a no-op diff in generated proto code, due to some difference in linting config. I am also curious why our linting workflow didn't catch this.

src/leapfrogai_sdk/proto/leapfrogai_sdk/completion/completion.proto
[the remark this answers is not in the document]
lemonprogis: Interesting, I'll take a look but I ran 'make gen-python'. Should there be more granularity added where one would only generate a specific module?

src/leapfrogai_sdk/audio/audio_pb2.py
lemonprogis: You are correct, I didn't have pre-commits setup on the repo yet. I reverted the initial commit, setup pre-commit, and then made the changes.

src/leapfrogai_sdk/proto/leapfrogai_sdk/completion/completion.proto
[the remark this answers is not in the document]
gphorvath: I'm not sure, but with the linter tweaks this looks better!

gphorvath (approved):

jalling97 (approved):
Thanks for pushing these fixes!

justinthelaw (approved):
@lemonprogis feel free to smash that merge button!

PR #1066, opened 2024-09-18
fix(whisper): fix whisper transcription of non-english audio

CollectiveUnicorn:
## Description
* The whisper backend was not correctly transcribing non-english audio into non-english text with or without the language flag. This fix ensures that the original language of the audio is what is outputted into the transcribed text.
* Adds more logging to the whisper backend
* Remove english as default and instead use the automatic language detection

### CHANGES

* The metadata is now set automatically if it exists to prevent issues where missing items would cause it to default to english.

## Checklist before merging


justinthelaw (approved):
LGTM!

Added an issue to eventually refactor some of Whisper back towards some of the SDK and Model Directory goals: https://github.com/defenseunicorns/leapfrogai/issues/1108

packages/whisper/main.py
```diff
@@ -62,26 +66,35 @@ def call_whisper(
...
+            updated = True
+
+        if metadata.inputlanguage:
+            input_language = metadata.inputlanguage
+            updated = True
+
+        if updated:
+            logger.info(
+                f"Updated metadata: Prompt='{prompt}', Temperature={temperature}, Input Language='{input_language}'"
+            )
```
jalling97, line 91: We could log these individually so we know which metadata fields are actually getting updated (as opposed to here where if only one of them gets updated then they are all logged) but I don't think it's really that big an issue so feel free to resolve this convo if you want.

tests/e2e/test_whisper.py
```diff
@@ -73,3 +73,57 @@ def is_english_or_punctuation(c):
     english_chars = [is_english_or_punctuation(c) for c in translation.text]
 
     assert all(english_chars), "Non-English characters have been returned"
+
+
+def test_non_english_transcription(client: OpenAI):
```
jalling97, line 78: Pedantic question: should each transcription be a separate test?

CollectiveUnicorn: They could be, but the purpose is basically just to check if the multi-lingual functions are working for transcription. Not necessarily that Russian and Arabic are both supported by transcription.

packages/whisper/main.py
CollectiveUnicorn: Added logging to each piece of metadata independent of one another.

tests/e2e/test_whisper.py
jalling97: That's fair!

jalling97 (approved):

justinthelaw (approved):

PR #1078, opened 2024-09-19
feat: 1049 add standard evaluation benchmarks to lfai evals

jalling97:
## Description

Adds baseline generation evaluation benchmarks to provide a known point of comparison. These benchmarks can show us when models we have implemented are not performing how we expect them to (based on reported eval benchmarks) and to better understand the impact of specific quantizations on certain models.

### BREAKING CHANGES

N/A

### CHANGES

* Refactors how evals get imported to make adding future evals easier
* Adds the MMLU and HumanEval benchmarks to `leapfrogai_evals`
* Adds a DeepEval compatible LLM class that relies on the LFAI API (not to be used as an LLM as a Judge)
* Updates the evals README
* Upgrades DeepEval to `v1.3.0`

## Related Issue

Relates to #1049 

## Checklist before merging

- [x] Tests, documentation, ADR added or updated as needed
- [x] Followed the [Contributor Guide Steps](https://github.com/defenseunicorns/leapfrogai/blob/main/.github/CONTRIBUTING.md)

PR #1090, opened 2024-09-20
feat(api): reranking backend integrated in with rag

CollectiveUnicorn:
## Description

* Adds reranking to RAG pipeline
* Adds RAG configuration endpoint when in dev mode
* Adds additional logging
* Refactors the pytest's `test_routes` api tests
* Alters default RAG values into two steps, retrieval and ranking. With the retrieval results being set to 100 after ranking the results are filtered down to the user specified k value. If reranking is not enabled, the user specified k results is returned from the retrieval step.
* Adds Zarf configs to enable dev mode.

## FlashRank Evals
* Rerank (`ms-marco-TinyBERT-L-2-v2`) + 100 top-k retrieval

```
Final Results:
INFO:root:Average Needle in a Haystack (NIAH) Retrieval: 1.0
INFO:root:Average Needle in a Haystack (NIAH) Response: 1.0
INFO:root:Average Correctness (GEval): 0.812
INFO:root:Average Answer Relevancy: 0.9590000000000001
INFO:root:Average Annotation Relevancy: 0.92
INFO:root:MMLU: 0.697979797979798
INFO:root:HumanEval: 0.96
INFO:root:Eval Execution Runtime (seconds): 3122.107043981552
```

* Rerank (`ms-marco-TinyBERT-L-2-v2`) + 10 top-k retrieval
```
Final Results:
INFO:root:Average Needle in a Haystack (NIAH) Retrieval: 1.0
INFO:root:Average Needle in a Haystack (NIAH) Response: 1.0
INFO:root:Average Correctness (GEval): 0.8140000000000001
INFO:root:Average Answer Relevancy: 0.9470000000000001
INFO:root:Average Annotation Relevancy: 0.92
INFO:root:MMLU: 0.697979797979798
INFO:root:HumanEval: 0.96
INFO:root:Eval Execution Runtime (seconds): 3119.72075009346
```
* No Rerank
```
Final Results:
INFO:root:Average Needle in a Haystack (NIAH) Retrieval: 1.0
INFO:root:Average Needle in a Haystack (NIAH) Response: 1.0
INFO:root:Average Correctness (GEval): 0.8299999999999998
INFO:root:Average Answer Relevancy: 0.9555128205128205
INFO:root:Average Annotation Relevancy: 0.92
INFO:root:MMLU: 0.697979797979798
INFO:root:HumanEval: 0.96
INFO:root:Eval Execution Runtime (seconds): 3112.184923171997
```
* [Baseline](https://github.com/defenseunicorns/leapfrogai/pull/1078#issuecomment-2372599711) - No Rerank
```
Final Results:
INFO:root:Average Needle in a Haystack (NIAH) Retrieval: 1.0
INFO:root:Average Needle in a Haystack (NIAH) Response: 1.0
INFO:root:Average Correctness (GEval): 0.81
INFO:root:Average Answer Relevancy: 0.9616666666666666
INFO:root:Average Annotation Relevancy: 0.92
INFO:root:MMLU: 0.695959595959596
INFO:root:HumanEval: 0.96
INFO:root:Eval Execution Runtime (seconds): 1961.5159051418304
```

## Checklist before merging


justinthelaw (approved):
The MVP implementation logic is sound to me - my only worry is a mix of a) our evals being skewed (too easy, too hard, for certain techniques, etc.) and, b) this re-ranking actually making our eval results worse with a noticable performance hit for the end-user. Not for this PR to solve, just my thoughts!

CollectiveUnicorn:
Yeah the way the eval results are atm makes things a bit hard to interpret. But performance I think isn't as different as it seems. The baseline was from @jalling97's machine vs mine. The actual time difference between the version without reranking and with reranking is between 7 and 10 seconds over a 51 minute test. Would be better if we got an average per request but I suppose that could be evaluated separately.

jalling97:
The `Annotation Relevancy` metric is what I was initially concerned with when I saw the results. I would have expected it to change (in either direction) after reranking was implemented. However, from what I can tell, the top_k chunks (with or without reranking) correspond to the same documents either way, resulting in no practical difference in annotations. Since the `Annotation Relevancy` metric isn't concerned with the order of the documents (just that the ones you expect are in the list of annotations),  the results are pretty much unchanged. 

This is good to note, as the next stage of evals will incorporate the chunk data thanks to #870, so we can better evaluate the rank of the chunks themselves. 

Overall I think this is a net gain and we'll have to see with the next wave of evals what practical difference the reranker has.

jalling97 (dismissed):

tests/utils/client.py
```diff
@@ -4,6 +4,10 @@
 import requests
 from requests import Response
 
+LEAPFROGAI_MODEL = os.getenv("LEAPFROGAI_MODEL", "llama-cpp-python")
+OPENAI_MODEL = "gpt-4o-mini"
```
justinthelaw, line 8: Similarly, this should use the `get_openai_model()` method on line 35.

tests/utils/client.py
```diff
@@ -4,6 +4,10 @@
...
+ROOT_URL = "https://leapfrogai-api.uds.dev"
```
justinthelaw, line 9: Similarly, this should use the `get_leapfrogai_api_url()` method on line 64.

justinthelaw (changes requested):

tests/utils/client.py
CollectiveUnicorn: Can't use the `get_leapfrogai_api_url` because that's the full url with `openai/v1` at the end

justinthelaw: Sorry, my bad! I meant `get_leapfrogai_api_url_base()`

CollectiveUnicorn: I actually think its already switched, I just need to get rid of this.

jalling97 (dismissed):

tests/integration/api/test_rag_files.py
```diff
@@ -1,9 +1,13 @@
...
+from typing import Optional
+
+import requests
 from openai.types.beta.threads.text import Text
 import pytest
 from tests.utils.data_path import data_path
 
-from tests.utils.client import client_config_factory
+from leapfrogai_api.typedef.rag.rag_types import ConfigurationPayload
+from tests.utils.client import ROOT_URL, client_config_factory
```
justinthelaw, line 10: This import should be changed over to the `get_leapfrogai_api_url_base()` method

justinthelaw (approved):

jalling97 (approved):

PR #1151, opened 2024-09-25
fix(ui): assistants chat

andrewrisse:
## Description
There was an issue with chatting with assistants where if you created a new one and went back to your previous assistant chat, the messages were gone.

Investigation revealed that there were some edge cases where server side fetched data was not in sync with client side store state.

This resulted in a large refactor to fetch data on the client side, instead of server side. This decision was made because the multiple pages depend on threads, files, and assistants and that data is modified in various places. Invalidating data at the root level whenever this data changed (triggering a load function that reloaded all threads, assistants, and files) would be too much overhead and network data. It's also complicated to keep the server fetched data in sync with the modified versions in our stores. It's simpler, cleaner, and less buggy to just fetch the data client side and manage it in the stores instead of trying to fetch it all server side, then update the stores.

I found some logical and testing issues during this refactor that had to be addressed for tests to pass, so there are several changes throughout, as well as new/modified tests.

Also fixes an issue where users could upload audio files to the file-managment page or assistants. Audio files are only allowed on the chat page without assistants.

## Checklist before merging


gregclark-defenseunicorns (approved):
Design Review passed

justinthelaw (approved):

